PR #24, opened 2017-10-18
feat: use --html-reporter- as a prefix for the cli options

j0tunn:


j0tunn:
@sipayRT @DudaGod @tormozz48

PR #63, opened 2018-03-17
feat: add ability to fit images in one row

sipayRT:
Add button to scale images with diff (functionality from gemini-gui by @reubenmoes)

before button click
image

after button click
image


lib/static/styles.css
```diff
@@ -69,24 +69,30 @@
...
 .report_show-only-errors .section_status_skipped {
     display: none;
 }
 
 .image-box__image {
     padding:  0 5px;
-    display: inline-block;
+    display: block;
     vertical-align: top;
+    flex-basis: 33.333%;
```
DudaGod: максимально точно

PR #138, opened 2018-08-13
feat: Sticky "Accept" button

victor-homyakov:
Accepting tall screenshots requires scrolling forth and back several times: scroll forth to compare the screenshots, and then scroll back to the "Accept" button. Sticky button minimizes scrolling.


sipayRT:
tnx for your PR!

PR #170, opened 2018-12-04
FIX: in hermione "plugins" is a top level property.

dooman87:


sipayRT:
hi

can you change your commit message to `docs: in hermione "plugins" is a top level property.`

PR #175, opened 2018-12-17
Cross plaform scripts and tests

Kreozot:
1. Added 'cross-env' to cross-platform run package scripts with enviroment variables.
2. Added 'path.join' constructions in tests to make them pass on different platforms.


Kreozot:
Is this ok now?

sipayRT:
sorry, but commit with tests it's not fix but `chore` :(

Kreozot:
Now? )

PR #180, opened 2019-01-17
fix: copyImageAsync uses non-promisified method

4eb0da:
It should use `fs.copyAsync` instead of `fs.copy`, because `fs.copy` expects callback (and it have not warnings about that)
Also found similair in hermione: https://github.com/gemini-testing/hermione/blob/4e6e7b99ebb39d470535d10805818f40ea6f46e1/lib/browser/commands/assert-view/capture-processors/update-refs.js#L5

### DIFF
--- a/lib/server-utils.js
+++ b/lib/server-utils.js
@@ -44,7 +44,7 @@ function createPath(kind, result, stateName) {
 
 function copyImageAsync(srcPath, destPath) {
     return makeDirFor(destPath)
-        .then(() => fs.copy(srcPath, destPath));
+        .then(() => fs.copyAsync(srcPath, destPath));
 }
 
 /**
--- a/test/lib/server-utils.js
+++ b/test/lib/server-utils.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const path = require('path');
+const fs = require('fs-extra');
 const utils = require('../../lib/server-utils');
 
 describe('server-utils', () => {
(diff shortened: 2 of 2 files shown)


lib/server-utils.js
```diff
@@ -44,7 +44,7 @@ function createPath(kind, result, stateName) {
 
 function copyImageAsync(srcPath, destPath) {
     return makeDirFor(destPath)
-        .then(() => fs.copy(srcPath, destPath));
```
j0tunn: https://github.com/jprichardson/node-fs-extra#sync-vs-async-vs-asyncawait

> Most methods are async by default. All async methods will return a promise if the callback isn't passed

4eb0da: In version 2.1.2 it returns nothing actually
https://github.com/gemini-testing/html-reporter/blob/master/package.json#L39
https://github.com/jprichardson/node-fs-extra/tree/v2.1.2#promises

j0tunn: Looks like v2.1.2 doesn't have `copyAsync` method. So the best solution is to update `fs-extra` package

4eb0da: It's provided by `bluebird` in several places, for example:
https://github.com/gemini-testing/html-reporter/blob/183541ee71cd2227d1d946d16c545af3b4bc7312/lib/plugin-adapter.js#L11

`copyAsync` already used in html-reporter:
https://github.com/gemini-testing/html-reporter/blob/12b332b2c7c5abc713b67229de802f3e470d56c8/lib/report-builder-factory/report-builder.js#L236

>  So the best solution is to update fs-extra package

Definitely. Buuut it's quite bigger task to achieve. One should upgrade 5 major versions of `fs-extra`, cleanup promisifyAll and tests
Moreover, better solution proposes the same version of the `fs-extra` in all tools: hermione, gemini and html-reporter

Maybe I can work on this for `html-reporter` later

j0tunn:
@sipayRT влейте, плз, и раскатите в следующей версии html-reporter'а

PR #187, opened 2019-01-30
feat(gui): ability to accept only opened tests

DudaGod:


lib/static/components/controls/gui-controls.js
```diff
@@ -25,10 +25,10 @@ class GuiControls extends Component {
         return actions.runFailedTests(failed);
     }
 
-    _acceptAll = () => {
+    _acceptOpens = () => {
```
kvmamich: `acceptOpened`?

PR #189, opened 2019-02-08
fix: filter browsers on render

Flackus:


lib/static/components/section/section-common.js
```diff
@@ -76,19 +84,11 @@ export class SectionCommon extends SectionBase {
...
-        } else {
-            suite = origSuite;
-        }
-
         return {
             expand,
             viewMode,
-            suite
+            suite: suites[ownProps.suiteId],
+            filteredBrowsers
```
CatWithApple: "filteredBrowsers" may be removed from this line if SectionCommon (lib/static/components/suites.js) will be created with "filteredBrowsers" as property. look at "filterByName".
there will be less confusion.

Flackus:
@DudaGod ping.

PR #192, opened 2019-02-12
fix: do not fail if one of the reports is empty

eGavr:
сс @DudaGod @rostik404 @DudaGod

### DIFF
--- a/test/lib/merge-reports/data-tree.js
+++ b/test/lib/merge-reports/data-tree.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
 const path = require('path');
 const fs = require('fs-extra');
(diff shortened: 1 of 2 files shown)


test/lib/merge-reports/data-tree.js
```diff
@@ -14,9 +13,7 @@ describe('lib/merge-reports/data-tree', () => {
         return DataTree.create(initialData, destPath);
     };
 
-    const mkData_ = (data = {}) => {
-        return _.defaults(data, {suites: [], skips: []});
-    };
+    const mkData_ = (data = {}) => data;
```
xrsd: Do you leave `mkData_`-function for the future modifications/extensions? Now it looks like unnecessary function.

eGavr: fixed

PR #206, opened 2019-03-12
feat(gui): do not use reuse by default

sipayRT:
BREAKING CHANGE: changed tests result reuse behavior. By default reuse will be disabled. Use "--reuse" option to enable it

DO NOT MERGE YET


CatWithApple:
> DO NOT MERGE YET
just use draft
https://github.blog/2019-02-14-introducing-draft-pull-requests/

DudaGod:
@sipayRT is it still relevant?

PR #217, opened 2019-04-02
feat: ability to add meta info extenders

DudaGod:


README.md
```diff
@@ -125,3 +125,23 @@ Adds item to html report as link:
...
+### addMetaInfoExtender
+
+Extend meta-info of each test using passed data:
+
+```js
+tool.htmlReporter.addMetaInfoExtender(name, value);
+```
+
+* **name** (required) `String` - name of meta info
+* **value** (required) `String|Number|Function` - value of meta info. If a function is passed, then `suite` and `extraItems` will be passed to its arguments.
```
kvmamich: It is unclear what is inside `extraItems`

lib/report-builder-factory/report-builder.js
```diff
@@ -104,8 +104,8 @@ module.exports = class ReportBuilder {
         return this;
     }
 
-    setExtraItems(extraItems) {
-        this._extraItems = extraItems;
+    setApiValues(values) {
```
kvmamich: Why is it called `setApiValues`? What API is this about?

DudaGod: it's about extra items and meta info extenders. I will think about more clearer name of this method

README.md
DudaGod: okay, I will add an example

lib/static/components/section/body/index.js
```diff
@@ -78,6 +78,16 @@ class Body extends Component {
         this.props.actions.toggleStateResult({browserId, suitePath, stateName, retryIndex, opened});
     }
 
+    getExtraMetaInfo = () => {
+        const {suite, apiValues: {extraItems, metaInfoExtenders}} = this.props;
+
+        return omitBy(mapValues(metaInfoExtenders, (extender) => {
+            const stringifiedFn = extender.startsWith('return') ? extender : `return ${extender}`;
+
+            return new Function(stringifiedFn)()(suite, extraItems);
```
sipayRT: how it works? You'll pass some 'bla' as a value for extender and here it will be `function() { return 'bla'}()(suite, extraItems)`?

lib/static/components/section/body/index.js
```diff
@@ -78,6 +78,16 @@ class Body extends Component {
         this.props.actions.toggleStateResult({browserId, suitePath, stateName, retryIndex, opened});
     }
 
+    getExtraMetaInfo = () => {
+        const {suite, apiValues: {extraItems, metaInfoExtenders}} = this.props;
+
+        return omitBy(mapValues(metaInfoExtenders, (extender) => {
+            const stringifiedFn = extender.startsWith('return') ? extender : `return ${extender}`;
```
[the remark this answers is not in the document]
DudaGod: Yes, of course.

For example your extender is:
```js
var fn = function() {
  return 'hello world';
}
```

I must stringified it in order to use inside react component. When I stringified it I get:
`"function() {\n  return 'hello world';\n}"`

Now I should transform it to normal function. For this I should use `eval` or `new Function`. I choose second variant. If you call `new Function(stringFn)` you get a wrapper function around passed string function. Therefore I should add `return` in order when I call wrapper I get function inside it and then I can pass args to it

lib/static/components/section/body/index.js
DudaGod: I describe how it works in my previous comment. And it will work only with functions. I fix docs and now extender can only be a function.

PR #220, opened 2019-04-09
fix: slow build image diff

CatWithApple:
Related PR:
https://github.com/gemini-testing/hermione/pull/397


CatWithApple:
rewriting the code

PR #221, opened 2019-04-09
fix(gui): idle status for suite if some of browsers has such status

sipayRT:
Fix case when suite has some test with IDLE status and all branch will be marked as idle, but others tests has final status (success, fail etc)

### DIFF
--- a/test/lib/common-utils.js
+++ b/test/lib/common-utils.js
@@ -1,0 +1,14 @@
+'use strict';
+
+const {determineStatus} = require('lib/common-utils');
+const {SUCCESS, IDLE} = require('lib/constants/test-statuses');
+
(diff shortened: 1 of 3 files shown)


test/lib/common-utils.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const {determineStatus} = require('lib/common-utils');
+const {SUCCESS, IDLE} = require('lib/constants/test-statuses');
+
+describe('common-utils', () => {
+    describe('determineStatus', () => {
+        it(`should not rewrite suite status with ${IDLE} if some test already has final status`, () => {
```
DudaGod: to my mind `... with idle` shoul be changed on `... to idle`

PR #233, opened 2019-06-21
feat: 'serverReady' event

SlNPacifist:
Added event when server is ready. Going to use this to profile time from process start to gui opening.


eGavr:
ok

PR #253, opened 2019-08-08
feat: add an ability to stop server directly from GUI

Hypnosphi:
This adds a "Stop server" button that sends a stop signal to the server, and disables all the actions that require server communication.

My usecase is to run `hermione gui` on CI (TeamCity), and then review & accept changes manually on agent. I need some way to signal that the review is finished and the artifacts can be saved.


Hypnosphi:
@sipayRT @TheNech @DudaGod can you please take a look?

sipayRT:
hi
will check it today

lib/static/gui.css
```diff
@@ -64,7 +64,7 @@
     min-width: 2vw;
     min-height: 2vh;
 }
-.button_blink {
+.button_loading {
```
Hypnosphi: Will undo

lib/static/modules/reducer.js
```diff
@@ -132,6 +132,9 @@ function reducer(state = getInitialState(compiledData), action) {
                 }
             };
         }
+        case actionNames.STOP_SERVER: {
+            return {...state, serverStopped: true};
```
sipayRT: you should set to `false` options as `running` and `processing`, because you can stop server while tests executing

test/unit/lib/static/components/controls/gui-controls.js
```diff
@@ -124,4 +141,22 @@ describe('<ControlButtons />', () => {
             });
         });
     });
+
+    describe('"Stop server" button', () => {
```
sipayRT: you need test on stop server while tests are already executing

lib/static/components/controls/gui-controls.js
```diff
@@ -33,27 +34,32 @@ class GuiControls extends Component {
...
                     <ControlButton
                         label="Retry failed tests"
-                        isDisabled={!failed.length || processing}
+                        isDisabled={!failed.length || processing || serverStopped}
                         handler={this._runFailedTests}
                     />
                     <ControlButton
                         label="Accept opened"
-                        isDisabled={!failed.length || processing}
+                        isDisabled={!failed.length || processing || serverStopped}
```
DudaGod: I suggest to refactor this piece of code. Let's create a function for this in which we get length of the first argument and add `|| processing || serverStopped`

Hypnosphi:
> Why you update package-lock.json?

`png-img@2.2.0` doesn't work with Node 12

lib/gui/server.js
```diff
@@ -64,6 +64,11 @@ exports.start = ({paths, hermione, guiApi, configs}) => {
         }
     });
 
+    server.post('/stop', (req, res) => {
+        res.sendStatus(200);
+        process.exit(0);
```
DudaGod: here we should stop also hermione process, otherwise the tests will still be executed

Hypnosphi: It should be stopped by `onExit` callback below, see [`signal-exit`](https://github.com/tapjs/signal-exit) docs:

> explicitly having `process.exit(code)` called.

DudaGod: I wanted to say that now there is a problem, that before completing the hermione process we should close browsers. Otherwise browsers are hanging. We will add api for this to hermione soon and then merge your PR.

You can reproduce it when run test with long timeout (30 sec for example) on local selenium server.

Hypnosphi: > We will add api for this to hermione soon

Hi, can I track the progress of this task anywhere?

DudaGod: > Hi, can I track the progress of this task anywhere?

Hello. Unfortunately no one from our team is doing this task now.

Hypnosphi: I think I found a way to do it. Can you please check again?

Hypnosphi:
@DudaGod is there anything else left to do here?

Hypnosphi:
merged as #397

Hypnosphi:
I'll take a look into making it a plugin, thank you

Hypnosphi:
@DudaGod what extension point would you recommend for that?

DudaGod:
> @DudaGod what extension point would you recommend for that?

I would recommend to use extension point with name `menu-bar`. Your button will be rendered inside special burger menu in header.

PR #266, opened 2019-09-24
fix: determine section status depending on filtered browsers

victor-homyakov:


lib/static/components/section/section-common.js
```diff
@@ -62,11 +64,19 @@ class SectionCommon extends Component {
...
         const fullTestName = suitePath.join(' ');
 
+        const browsersToShow = browsers
+            .filter(browser => shouldBrowserBeShown({browser, fullTestName, filteredBrowsers, errorGroupTests}));
+
+        // determine section status depending on filtered browsers;
+        // mark empty section as successful
+        const status = browsersToShow.length > 0 ?
+            determineStatus(browsersToShow.map(({result: {status}}) => status)) :
+            SUCCESS;
```
kvmamich: Why do we need status for empty section?

victor-homyakov: The component's code right now has no checks or constraints related to its emptiness. So to be sure I supported both cases: empty and non-empty section.

victor-homyakov:
@xrsd ping?

Issue #247, opened 2019-07-23
Suggestion: Add favicon notification when test passed


R1ZEN:
Favicon notification helps to understand when the test passed looking on the tab bar.

![image](https://user-images.githubusercontent.com/21066057/61717402-a7ce5780-ad69-11e9-84ec-a99af2609a01.png)

Example: http://igorprado.com/favicon-notification/

DudaGod:
Hello.
The link - http://igorprado.com/favicon-notification/ you sent no longer responds. But anyway I understand your suggestion. Can you send PR with implementation of this feature? )

R1ZEN:
Hi! When i try to install reporter `node_modules`, i catch the issue with module `png-img` on `nodejs 14`, also when `node_modules` are fully installed and trying to start `html-reporter`, got the error:
```
internal/modules/cjs/loader.js:638
    throw err;
    ^

Error: Cannot find module 'webpack-cli/bin/config-yargs'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/Users/pavel/Projects/opensource/html-reporter/node_modules/webpack-dev-server/bin/webpack-dev-server.js:65:1)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
```

DudaGod:
I'm fixing it right now.

DudaGod:
I fixed it, now it should not be reproduced.

R1ZEN:
Good! I'll try to implement this feature and send PR

DudaGod:
It was merged here - https://github.com/gemini-testing/html-reporter/pull/350.

PR #321, opened 2020-05-11
feat: implement favicon-notification

R1ZEN:
This PR add possibility to check tests status in browser tab. This feature potentially improves user experience, because the user no longer needs to go to the GUI tab or console to understand, tests are running or not.
Suggestion: https://github.com/gemini-testing/html-reporter/issues/247 and Screenshot:
image

### DIFF
--- a/lib/static/components/favicon-status.js
+++ b/lib/static/components/favicon-status.js
@@ -3,53 +3,78 @@
 import {PureComponent} from 'react';
 import {connect} from 'react-redux';
 import PropTypes from 'prop-types';
+import {IDLE, RUNNING, SUCCESS, FAIL} from '../../constants/test-statuses';
 
-const DEFAULT = 'default';
-const SUCCESS = 'success';
-const FAILED = 'fail';
-const RUNNING = 'running';
+const GREY = '#ccc';
+const GREEN = '#038035';
+const RED = '#c00';
+const YELLOW = '#e7d700';
 
 class FaviconStatus extends PureComponent {
     static propTypes = {
-        status: PropTypes.string
+        running: PropTypes.bool,
+        isFailedSuiteExist: PropTypes.bool
     }
 
-    link = null;
+    constructor(props) {
+        super(props);
+        this.notRunning = true;
+        this.link = document.head.querySelector('link[rel="icon"]');
+    }
 
     getFaviconURL(color) {
         const canvas = document.createElement('canvas');
         const diameter = 32;
+        const circeSize = diameter / 3;
 
         canvas.height = diameter;
         canvas.width = diameter;
         const ctx = canvas.getContext('2d');
 
         ctx.beginPath();
-        ctx.arc(diameter / 2, diameter / 2, diameter / 3, 0, 2 * Math.PI);
+        ctx.arc(diameter / 2, diameter / 2, circeSize, 0, 2 * Math.PI);
         ctx.fillStyle = color;
         ctx.fill();
 
         return canvas.toDataURL('image/png');
     }
 
-    render() {
-        const {status} = this.props;
-        if (!this.link) {
-            this.link = document.head.querySelector('link[rel="icon"]');
+    changeIconColor(color) {
+        this.link.setAttribute('href', this.getFaviconURL(color));
+    }
+
+    getSuiteStatus = () => {
+        const {running, isFailedSuiteExist} = this.props;
+
+        if (running) {
+            this.notRunning = false;
+            return RUNNING;
         }
 
-        switch (status) {
+        if (this.notRunning) {
+            return IDLE;
+        }
+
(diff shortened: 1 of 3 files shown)


lib/static/components/favicon-status.js
```diff
@@ -0,0 +1,76 @@
...
+        return null;
+    }
+}
+
+export default connect(
+    ({reporter: {running, suiteIds}}) => {
+        let status;
+        if (running) {
+            status = RUNNING;
+        } else if (!suiteIds.all.length && !suiteIds.failed.length) {
```
DudaGod, line 64: It does not work correctly because when we open gui we already have suites which is not running yet. And in this case green circle rendered in favicon, but should be grey.

lib/static/components/favicon-status.js
```diff
@@ -0,0 +1,76 @@
...
+class FaviconStatus extends PureComponent {
+    static propTypes = {
+        status: PropTypes.string
+    }
+
+    link = null;
+
+    getFaviconURL(color) {
+        const canvas = document.createElement('canvas');
+        const diameter = 32;
```
DudaGod, line 21: maybe we should get ability to control this value from plugin config?

lib/static/components/favicon-status.js
```diff
@@ -0,0 +1,76 @@
...
+
+        canvas.height = diameter;
+        canvas.width = diameter;
+        const ctx = canvas.getContext('2d');
+
+        ctx.beginPath();
+        ctx.arc(diameter / 2, diameter / 2, diameter / 3, 0, 2 * Math.PI);
```
DudaGod, line 28: If diameter is - `32` why radius - `32 / 3`?

lib/static/components/favicon-status.js
```diff
@@ -0,0 +1,76 @@
...
+            status = FAILED;
+        } else if (!suiteIds.failed.length) {
+            status = SUCCESS;
+        }
+
+        return {
+            status
+        };
+    }
+)(FaviconStatus);
```
DudaGod, line 76: I suggest covering this component with tests

lib/static/components/favicon-status.js
R1ZEN: In my opinion, `32 / 2` is a large circle size for the `32x32` icon) but if you think it is more preferable I will fix it)

lib/static/components/favicon-status.js
R1ZEN: Not sure that will be useful for configuration, can you imagine case when you need configure size of favicon?

lib/static/components/favicon-status.js
```diff
@@ -0,0 +1,76 @@
+'use-strict';
+
+import {PureComponent} from 'react';
+import {connect} from 'react-redux';
+import PropTypes from 'prop-types';
+
+const DEFAULT = 'default';
```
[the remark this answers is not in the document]
R1ZEN: Yep, it should be `idle` status

lib/static/components/favicon-status.js
```diff
@@ -0,0 +1,76 @@
...
+const SUCCESS = 'success';
+const FAILED = 'fail';
+const RUNNING = 'running';
+
+class FaviconStatus extends PureComponent {
+    static propTypes = {
+        status: PropTypes.string
```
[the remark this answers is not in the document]
R1ZEN: Because `status` comes from `connect` i'll be sure that the component always has `status`

lib/static/components/favicon-status.js
```diff
@@ -0,0 +1,76 @@
...
+            status = DEFAULT;
+        } else if (suiteIds.failed.length) {
+            status = FAILED;
+        } else if (!suiteIds.failed.length) {
+            status = SUCCESS;
+        }
+
+        return {
+            status
+        };
```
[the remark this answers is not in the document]
R1ZEN: I agree with that, but I didn't find `status` in the store. I found the calculation of `status` based on the data from the store inside the `Run` component.

lib/static/components/favicon-status.js
DudaGod: Perhaps you're right, let's at least put it into constants

lib/static/components/favicon-status.js
DudaGod: I have not tried changing these sizes so I don't know how it looks like with radius `32 / 2`. I don't understand why the radius is so oddly calculated.

lib/static/components/favicon-status.js
R1ZEN: Fixed that logic

DudaGod:
I implemented this feature a little differently here - https://github.com/gemini-testing/html-reporter/pull/350.
So I close it.  In any case, thank you very much.

PR #326, opened 2020-05-20
fix(gui): do not show silently skipped tests in tree

sipayRT:


lib/gui/tool-runner/index.js
```diff
@@ -200,6 +200,10 @@ module.exports = class ToolRunner {
         await this._fillTestsTree();
     }
 
+    _isSilentlySkipped({silentSkip, parent}) {
+        return silentSkip || parent && this._isSilentlySkipped(parent);
```
DudaGod, line 204: 🔥

PR #351, opened 2020-09-14
fix: speed up render components

DudaGod:
### What is done:
- component `State`:
  - move call action `toggleStateResult` from `constructor` to `componentDidMount`, gives speed up on 80ms;
  - do not call action `toggleStateResult` for images without `stateName`;
  - do not call action `toggleStateResult` with the same data as in redux store;
  - do not wrap action `triggerViewChanges` on `triggerViewChanges` which causes slowdown on 325 ms;
- component `Body`:
  - move call action `changeTestRetry` from `constructor` to `componentDidMount`, gives speed up on 5-10ms;
  - do not call action `changeTestRetry` with the same data as in redux store;
- component `StateError`:
  - do not parse html string for `hint` field if it is not visible yet (not opened by default), gives speed up on 10-20ms;
  - do not render `Screenshot` component inside `Page screenshot` section if it is not visible yet;
- component `Description`:
  - do not transform text to markdown it component is not visible yet (not opened by default), gives speed up on 10-20ms;
- component `MetaInfo`:
  - do not render all invisible data, hide behind a click;
- componen `SectionCommon`:
  - wrap child suite to lazily render component too;
  - use ability to reset state for already visible suite components, gives speed up on change expand to `all` from 20-30s to 2-3s. Related PR - https://github.com/gemini-testing/react-lazily-render/pull/2

- update dependencies: react, react-dom, redux, react-redux, ...


package.json
```diff
@@ -118,16 +119,15 @@
...
-    "react-redux": "^5.0.6",
-    "redux": "^3.7.2",
+    "react-dom": "^16.13.1",
+    "react-redux": "^7.2.1",
+    "redux": "^4.0.5",
     "redux-logger": "^3.0.6",
     "redux-mock-store": "^1.5.1",
     "redux-thunk": "^2.2.0",
     "rimraf": "^2.6.3",
-    "semantic-ui-react": "^0.83.0",
```
sipayRT, line 130: он у нас больше не используется?

test/unit/lib/static/components/section/section-common.js
```diff
@@ -160,11 +160,23 @@ describe('<SectionCommon/>', () => {
                     {tree, view: {lazyLoadOffset: 100500}}
                 );
                 const lazy = component.find(LazilyRender);
-                lazy.invoke('onRender')();
+                const section = lazy.invoke('children')(false);
+
+                assert.isNull(section);
+            });
+
+            it('should render section section if it is inside viewport', () => {
```
sipayRT, line 168: section section?

sipayRT (approved):

package.json
DudaGod: Используется, только он почему-то был определен в dev-dependencies. А должен быть в dependencies. Я его обновил и перенес.

PR #379, opened 2021-05-24
fix(tinder mode): filter images by selected browser

DudaGod:


test/unit/lib/static/modules/selectors/tree.js
```diff
@@ -85,6 +85,33 @@ describe('tree selectors', () => {
             const imagesByStateName = getAcceptableImagesByStateName(state);
             assert.deepEqual(Object.keys(imagesByStateName), ['b1 first', 'b2 second']);
         });
+
+        it('should return images from browser which matched to browser name and version from filter', () => {
```
rostik404, line 89: should return images for the browser which is passed in the filter.

rostik404 (approved):

PR #382, opened 2021-06-03
Add 'failed tests' selector in plugins

Aracturat:
I need to get all failed tests with metadata in my plugin.
For this reason, I want to reuse redux reselect selectors.

This PR add these selectors to plugins and extend failedTests selector.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ directory.
   The routes then can be called from the plugin React components defined in the `plugin.js`. For convenience the plugin name is always passed with options when function- or array-returning form is used to export plugin as the function options property `pluginName`:
 
   ```js
-  export default ['react', 'axios', function(React, axios, {pluginName, pluginConfig, actions}) {
+  export default ['react', 'axios', function(React, axios, {pluginName, pluginConfig, actions, selectors}) {
       class PluginComponent extends React.Component {
           // ... somewhere inside the component ...
           const result = await axios.get(`/plugin-routes/${pluginName}/plugin-route`);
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -270,7 +270,7 @@ directory.
   }
   ```
 
-  In the example you can also see another convenience property `actions` with all the html-reporter **Redux** actions.
+  In the example you can also see another convenient properties `actions` and `selectors` with all the html-reporter **Redux** actions and selectors.
```
DudaGod, line 273: Maybe let's identify that `selectors` it's a memoized selectors which created using `reselect` library?

lib/static/modules/selectors/tree.js
```diff
@@ -34,7 +34,7 @@ export const getFailedTests = createSelector(
             const browserId = result.parentId;
             const {parentId: testName, name: browserName} = browsers[browserId];
 
-            return {testName, browserName};
+            return {testName, browserName, metaInfo: result.metaInfo};
```
DudaGod, line 37: In order to not remove `metaInfo` which is not used now inside html-reporter we should write a test.

lib/static/modules/selectors/index.js
```diff
@@ -0,0 +1,5 @@
+// There is a bug https://github.com/babel/babel-eslint/issues/167
```
DudaGod, line 1: It looks like that the problem described in this issue is already fixed am I right? Maybe we should update `babel-eslint`?

DudaGod (commented):
Let's add tests and I will merge it.

PR #383, opened 2021-06-04
feat(plugins): add menu-bar extension point

Aracturat:
Add new extension point to extend menu bar (other way is extraItems property).


lib/static/components/controls/menu-bar.js
```diff
@@ -20,15 +22,20 @@ class MenuBar extends Component {
     render() {
         const {extraItems} = this.props;
 
-        if (isEmpty(extraItems)) {
+        const hasMenuBarPlugins = plugins.getLoadedConfigs()
+            .some(e => e.point === 'menu-bar');
```
DudaGod, line 26: let's define available extension points inside separate file here - https://github.com/gemini-testing/html-reporter/tree/master/lib/constants. And then use them here and where specify `result` extension point.

DudaGod (commented):
Let's add tests for menu-bar component and I will merge it.

lib/static/components/controls/menu-bar.js
Aracturat: Done

Aracturat:
@DudaGod Added tests, please review and merge

Aracturat:
@DudaGod Added tests, please review and merge

PR #385, opened 2021-06-08
feat(plugins): add plugin configuration

Aracturat:
I need to pass some configuration to my plugin.
This PR adds it

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,8 @@ directory.
           name: 'plugin-name',
           component: 'PluginReactComponentName',
           point: 'extension-point-name',
-          position: 'wrap'
+          position: 'wrap',
+          config: { param: 'value'}
       },
       {
           name: 'plugin-name',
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -270,7 +271,7 @@ directory.
   }
   ```
 
-  In the example you can also see another convenience property `actions` with all the html-reporter **Redux** actions.
+  In the example you can also see another convenient properties `actions` and `pluginConfig` with all the html-reporter **Redux** actions and plugin configuration.
```
DudaGod, line 274: Maybe you should not try to write a challenging sentence? You can simply break it on points:
``` md
text-1:
- text-2
- text-3
```

README.md
```diff
@@ -184,6 +184,7 @@ directory.
       * `wrap` - to wrap the extension point UI
       * `before` - to place the component before the extension point
       * `after` - to place the component after the extension point
+  * **config** (optional) - plugin configuration
```
DudaGod, line 187: Let's add this field to the example above. And to my mind we should write available type. Like:
```
* **config** (optional) `Object` - ...
```

You can also describe other fields in the same way.

DudaGod (commented):
Looks good, let's write tests.

PR #419, opened 2022-03-31
fix(remove-unused-screens): bugs

DudaGod:


DudaGod:
This code reused from cli command of remove dumps.

PR #540, opened 2024-04-01
docs: proper default value for path option

miripiruni:
Changes in docs according to https://github.com/miripiruni/html-reporter/blob/master/lib/config/index.ts#L150


KuznetsovRoman (changes requested):
https://www.conventionalcommits.org/en/v1.0.0/ suggests to use `docs` tag in commit message:
```bash
git commit --amend -m "docs: proper default value for path option" && git push -f
```

KuznetsovRoman (approved):

PR #541, opened 2024-04-03
feat(html-reporter): add testplane support

KuznetsovRoman:
Merging straight to master

## What is done
- add native testplane@0 and testplane@1 support
- customize help messages (e.g. it suggests to open `gui` like `npx testplane gui` if it was launched with `npx testplane`)
- add testplane html-reporter plugins support
- add testplane custom gui support (with doubling `testplane` property in addition to `hermione`)

## TBD
- import from "testplane" instead of "hermione"


index.ts
```diff
@@ -0,0 +1,132 @@
...
+import {formatTestResult, logPathToHtmlReport, logError, getExpectedCacheKey} from './lib/server-utils';
+import {SqliteClient} from './lib/sqlite-client';
+import {HtmlReporterApi, ReporterOptions, TestSpecByPath} from './lib/types';
+import {createWorkers, CreateWorkersRunner} from './lib/workers/create-workers';
+import {SqliteImageStore} from './lib/image-store';
+import {Cache} from './lib/cache';
+import {ImagesInfoSaver} from './lib/images-info-saver';
+import {getStatus} from './lib/test-adapter/testplane';
+
+export default (testplane: Testplane, opts: Partial<ReporterOptions>): void => {
```
KuznetsovRoman, line 23: We **can** use `export default` for testplane

hermione.ts
```diff
@@ -1,131 +1,3 @@
...
-                }).catch(reject));
-            });
-        });
-
-        hermione.on(hermione.events.RUNNER_END, () => {
-            return Promise.all(promises).then(() => resolve(), reject);
-        });
-    });
-}
+export = pluginHandler;
```
KuznetsovRoman, line 3: Export default export of `./index.ts` as `module.exports`, because old versions of hermione do not support `export default`

lib/server-utils.ts
```diff
@@ -242,11 +243,11 @@ export function getDataForStaticFile(htmlReporter: HtmlReporter, pluginConfig: R
     };
 }
 
-export async function initializeCustomGui(hermione: Hermione, {customGui}: ReporterConfig): Promise<void> {
+export async function initializeCustomGui(testplane: Testplane, {customGui}: ReporterConfig): Promise<void> {
     await Promise.all(
         _(customGui)
             .flatMap<CustomGuiItem>(_.identity)
-            .map((ctx) => ctx.initialize?.({hermione, ctx}))
+            .map((ctx) => ctx.initialize?.({testplane, hermione: testplane, ctx}))
```
KuznetsovRoman, line 250: `testplane` and `hermione` props are `Object.is` copies for backward compatibility

lib/server-utils.ts
```diff
@@ -257,12 +258,12 @@ export interface CustomGuiActionPayload {
...
 }
 
-export async function runCustomGuiAction(hermione: Hermione, {customGui}: ReporterConfig, payload: CustomGuiActionPayload): Promise<void> {
+export async function runCustomGuiAction(testplane: Testplane, {customGui}: ReporterConfig, payload: CustomGuiActionPayload): Promise<void> {
     const {sectionName, groupIndex, controlIndex} = payload;
     const ctx = customGui[sectionName][groupIndex];
     const control = ctx.controls[controlIndex];
 
-    await ctx.action({hermione, control, ctx});
+    await ctx.action({testplane, hermione: testplane, control, ctx});
```
KuznetsovRoman, line 266: `testplane` and `hermione` props are `Object.is` copies for backward compatibility

lib/static/modules/load-plugin.js
```diff
@@ -105,7 +105,10 @@ async function initPlugin(plugin, pluginName, pluginConfig) {
 // Actual plugin is passed to __hermione_html_reporter_register_plugin__ somewhere in the
 // plugin code
 function executePluginCode(code) {
-    const exec = new Function(`function __hermione_html_reporter_register_plugin__(p) {return p;}; return ${code};`);
+    const getRegisterFn = tool => `function __${tool}_html_reporter_register_plugin__(p) {return p;};`;
+
+    const exec = new Function(`${getRegisterFn('testplane')} ${getRegisterFn('hermione')} return ${code};`);
+
```
KuznetsovRoman, line 111: Add support for testplane html reporter plugins

lib/types.ts
```diff
@@ -8,12 +8,15 @@ declare module 'tmp' {
     export const tmpdir: string;
 }
 
-export {Suite as HermioneSuite} from 'hermione';
+export {Suite as TestplaneSuite} from 'hermione';
 
-export interface HermioneTestResult extends HermioneTestResultOriginal {
+export interface HermioneTestResult extends TestResult {
```
KuznetsovRoman, line 13: Not a part of public API, no JSDoc deprecated's is needed

lib/types.ts
```diff
@@ -8,12 +8,15 @@ declare module 'tmp' {
...
     timestamp?: number;
 }
 
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface TestplaneTestResult extends HermioneTestResult {}
```
KuznetsovRoman, line 18: These two are separated, so we could add something to `TestplaneTestResult` while it still does not exist on `HermioneTestResult`

lib/types.ts
```diff
@@ -155,8 +158,23 @@ export interface PluginDescription {
...
     type: string;
     controls: {label: string; value: string;}[];
-    initialize?: (data: {hermione: Hermione, ctx: object}) => void | Promise<void>;
-    action: (data: {hermione: Hermione, ctx: object, control: object}) => void | Promise<void>;
+    initialize?: (data: {
+        testplane: Testplane,
+        /**
+         * @deprecated Use `testplane` instead
+         */
+        hermione: Testplane,
```
KuznetsovRoman, line 166: Add some JSDoc deprecated comments

test/unit/index.js
```diff
@@ -56,12 +56,12 @@ describe('lib/hermione', () => {
...
     }).HtmlReporter;
 
-    const runHtmlReporter = proxyquire('../../hermione', {
+    const runHtmlReporter = proxyquire('../../', {
         './lib/sqlite-client': {SqliteClient},
         './lib/server-utils': utils,
         './lib/report-builder/static': {StaticReportBuilder},
         './lib/plugin-api': {HtmlReporter}
-    });
+    }).default;
```
KuznetsovRoman, line 64: use default-exported value here

test/unit/lib/gui/tool-runner/index.js
```diff
@@ -40,44 +40,44 @@ describe('lib/gui/tool-runner/index', () => {
...
-        return ToolGuiReporter.create(opts.paths, hermione, configs);
+        return ToolGuiReporter.create(opts.paths, testplane, configs);
     };
 
     beforeEach(() => {
-        hermione = stubTool();
+        testplane = stubTool();
 
-        createTestRunner = sinon.stub();
+        createTestRunner = sandbox.stub();
```
KuznetsovRoman, line 73: replaced `sinon` with `sandbox` in the file, because some other tests (in another files) were failing without it

index.ts
shadowusr, line 1: Why `index.ts`? I think it should be `testplane.ts`. I strongly believe that importing from html-reporter shouldn't return testplane-related objects right away. We may choose to support many other tools in the future besides testplane and having testplane as main tool will be awkward. And removing it will be breaking change.

Ideally html-reporter should be tool-agnostic in every place, except a couple of adapters, IMO.

gemini.js
shadowusr, line 1: As far as I remember we've discussed a while ago that this file can be deleted altogether (since it already is throwing an error, after removing it, just error message will change for the end users and hence this is not a breaking change). Not critical to me though.

lib/constants/tool-names.ts
```diff
@@ -1,4 +1,5 @@
 export enum ToolName {
+    Testplane = 'testplane',
     Hermione = 'hermione',
```
shadowusr, line 3: Will `Hermione` tool name still be used after this PR? As far as I can see, you've removed `hermione` from everywhere, maybe we can remove it from this place as well?

lib/static/modules/load-plugin.js
shadowusr: This will work if both are present, right?

package.json
```diff
@@ -1,11 +1,12 @@
 {
   "name": "html-reporter",
   "version": "9.17.2",
-  "description": "Plugin for hermione which is intended to aggregate the results of tests running into html report",
+  "description": "Plugin for Testplane (ex-hermione) which is intended to aggregate the results of tests running into html report",
   "files": [
     "build"
   ],
   "exports": {
+    ".": "./build/index.js",
```
shadowusr, line 9: Here as well I am strongly against making testplane a default export. I think it should be testplane, for future extensibility to another tools. If you don't agree, we can discuss this with the team.

package.json
```diff
@@ -1,11 +1,12 @@
 {
   "name": "html-reporter",
   "version": "9.17.2",
-  "description": "Plugin for hermione which is intended to aggregate the results of tests running into html report",
+  "description": "Plugin for Testplane (ex-hermione) which is intended to aggregate the results of tests running into html report",
```
shadowusr, line 4: I'd change it to something more universal, like:
```
Html-reporter and GUI for viewing and managing results of a tests run. Currently supports Testplane and Hermione.
```
Not critical to me.

shadowusr (approved):
Looks good to me except making testplane a default export from the package, I am strongly against that.

PR #542, opened 2024-04-04
fix: dont require hermione dependency

KuznetsovRoman:
## What is done
- Add `testplane` peer dependency
- Mark `hermione` and `testplane` peer dependencies as `optional`
- Removed `hermione` usage from the project (`testplane` is now used)


package.json
```diff
@@ -59,7 +59,16 @@
...
+    "hermione": ">=6.0.0",
+    "testplane": "*"
+  },
+  "peerDependenciesMeta": {
+    "hermione": {
+      "optional": true
+    },
+    "testplane": {
+      "optional": true
+    }
```
KuznetsovRoman, line 71: The rest of the PR is just renaming

sipayRT (approved):

PR #544, opened 2024-04-25
feat: add error snippets support

KuznetsovRoman:
## What is done
- add ansi colors support to error message

image

- add error snippets support

image

image


lib/static/components/details.jsx
```diff
@@ -10,7 +10,8 @@ export default class Details extends Component {
         title: PropTypes.oneOfType([PropTypes.element, PropTypes.string]).isRequired,
         content: PropTypes.oneOfType([PropTypes.func, PropTypes.string, PropTypes.element, PropTypes.array]).isRequired,
         extendClassNames: PropTypes.oneOfType([PropTypes.array, PropTypes.string]),
-        onClick: PropTypes.func
+        onClick: PropTypes.func,
+        asHtml: PropTypes.bool,
```
KuznetsovRoman, line 14: Ability to pass raw html is needed in order to display our html, created from ansi

lib/static/components/state/state-error.jsx
```diff
@@ -6,13 +6,23 @@ import {bindActionCreators} from 'redux';
...
 import Details from '../details';
 import {ERROR_TITLE_TEXT_LENGTH} from '../../../constants/errors';
 import {isAssertViewError, isImageDiffError, isNoRefImageError} from '../../../common-utils';
 
+ansiHtml.setColors({
+    reset: ["#", "#"],
+    cyan: "ff6188",
+    yellow: "5cb008",
+    magenta: "8e81cd",
+    green: "aa8720",
```
KuznetsovRoman, line 23: Those are not cyan, yellow, magenta and green colors.
Just picked some with more contrast

lib/static/components/state/state-error.jsx
```diff
@@ -6,13 +6,23 @@ import {bindActionCreators} from 'redux';
...
+import ansiHtml from "ansi-html-community";
 import * as actions from '../../modules/actions';
 import ResizedScreenshot from './screenshot/resized';
 import ErrorDetails from './error-details';
 import Details from '../details';
 import {ERROR_TITLE_TEXT_LENGTH} from '../../../constants/errors';
 import {isAssertViewError, isImageDiffError, isNoRefImageError} from '../../../common-utils';
 
+ansiHtml.setColors({
+    reset: ["#", "#"],
```
KuznetsovRoman, line 19: hack to "not set bgcolor and text color"
Else frame background would be black.

lib/static/components/state/state-error.jsx
```diff
@@ -46,6 +56,24 @@ class StateError extends Component {
         return null;
     }
 
+    _tryToImproveSnippet(snippet) {
```
KuznetsovRoman, line 59: By default, `_errorToElements` slices our string into lines, and picks first one to be a title.

So at first it looked like this:

image

Here i am trying to pull out "filename" and use it as first line, so it would became title:

image

lib/static/components/state/state-error.jsx
```diff
@@ -46,6 +56,24 @@ class StateError extends Component {
...
+        const filePrefix = "| // ";
+        const filePrefixBeginIndex = snippet.indexOf(filePrefix);
+        const filePrefixEndIndex = snippet.indexOf("\x1B", filePrefixBeginIndex + filePrefix.length);
+
+        if (filePrefixBeginIndex === -1 || filePrefixEndIndex === -1) {
+            return snippet;
+        }
```
KuznetsovRoman, line 66: We are trying to find it by assuming our first line is "| // <file-name>" ended with control symbol (to stop painting text with gray, in this case)

Looks fragile, but i dont know any other option

lib/static/components/state/state-error.jsx
```diff
@@ -56,8 +84,17 @@ class StateError extends Component {
...
+                if (key === "snippet") {
+                    value = this._tryToImproveSnippet(value);
+                }
+
                 if (value.match(/\n/)) {
                     [titleText, ...content] = value.split('\n');
+
+                    if (!content[0].length) {
+                        content = content.slice(1);
+                    }
```
KuznetsovRoman, line 96: `Expect`-s have extra empty line in error message, which could be removed:

### Before
image


### After
image

test/unit/lib/static/components/state/state-error.jsx
```diff
@@ -101,4 +101,62 @@ describe('<StateError/> component', () => {
...
+            const error = {snippet: `\x1B[90m   . | // /some-file-path.js\x1B[39m")}
+                . |
+                9 | some line
+               10 | other line
+            `};
+
+            const component = mkStateErrorComponent({result: {error}});
+            component.find('.details__summary').last().simulate('click');
+
+            assert.isTrue(component.find('.details__content').html().startsWith('<div class="details__content"><pre>'));
```
KuznetsovRoman, line 116: For some reason, we cant use `.find` to get into this `<pre>` node. I think it is because i used `dangerouslySetInnerHTML`

lib/common-utils.ts
```diff
@@ -258,3 +258,99 @@ export const isImageBufferData = (imageData: ImageBuffer | ImageFile | ImageBase
...
+        const columnIndex = e.stack.indexOf(':');
+
+        if (columnIndex !== -1) {
+            errorName = e.stack.slice(0, columnIndex);
+        } else {
+            errorName = e.stack.slice(0, e.stack.indexOf('\n'));
+        }
+    }
+
+    if (!errorName) {
```
KuznetsovRoman, line 290: If we couldn't, use default value

lib/common-utils.ts
```diff
@@ -258,3 +258,99 @@ export const isImageBufferData = (imageData: ImageBuffer | ImageFile | ImageBase
...
+    return e.stack.slice(errorMessageEndsStackIndex);
+};
+
+const cloneError = <T extends Error>(error: T): T => {
+    const originalProperties = ['name', 'message', 'stack'] as Array<keyof Error>;
+    const clonedError = new Error(error.message) as T;
+
+    originalProperties.forEach(property => {
+        delete clonedError[property];
+    });
```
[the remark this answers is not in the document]
sipayRT: why is the correct order important here?

lib/common-utils.ts
```diff
@@ -258,3 +258,99 @@ export const isImageBufferData = (imageData: ImageBuffer | ImageFile | ImageBase
...
+            error.snippet
+        ].join('\n');
+
+        return clonedError;
+    }
+
+    const grayBegin = '\x1B[90m';
+    const grayEnd = '\x1B[39m';
+
+    clonedError.stack = [
```
sipayRT, line 349: it looks like duplicated code. it can be rewritten as

```javascript
const mergedStack = [getErrorTitle(error), error.snippet];

if (error.stack) {
    const grayBegin = '\x1B[90m';
    const grayEnd = '\x1B[39m';

    mergedStack.push(
        grayBegin + getErrorRawStackFrames(error as Error & { stack: string }) + grayEnd
    );
}

return mergedStack.join('\n');
```

also, we can add `colorCode` function for more readable code:

```javascript
function wrapWithAnsiColor(code, color = '90m') {
    return `\x1B[${color}${code}\x1B[39m`;
}
//...
if (error.stack) {
    const rawStackFrames = getErrorRawStackFrames(error as Error & { stack: string });

    mergedStack.push(
        wrapWithAnsiColor(rawStackFrames)
    );
}
```

sipayRT (approved):

lib/common-utils.ts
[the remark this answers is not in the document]
KuznetsovRoman: It matters here: https://github.com/gemini-testing/html-reporter/blob/master/lib/static/components/state/state-error.jsx#L49-L50

We want to se `name`, `message` and `stack` in our report

lib/common-utils.ts
KuznetsovRoman: jfyi: `m` is not a part of the color code in ansi.

Exampels:
- `\033[9;31;40m` (`[9` - Strikethrough, `31` - text color, `40` - bg color): strike-through red text with black bg
- `\033[1;37;42m` (`[1` - Semibold, `37` - text color, `42` - bg color): semibold white text with green bg
- `\033[0m` (`[0` - normal text, ... - reset text color, ... - reset bg color): removes all styling

PR #548, opened 2024-05-23
fix: use temp path to save diff image

shtyr-maxim:
This pull request fixes the error of storing diff images in the project's working directory.

## Expected behavior

The difference image is saved in `$TMPDIR/...` path before  copying to `reportPath`

## Actual behavior

The difference image is saved in `$PWD/images/...` before copying to `reportPath`


sipayRT:
hi

thx for your PR. can you, pls, provide an example to reproduce this bug?

shtyr-maxim:
Hi! Here is my minimal environment to reproduce these issue:
[html-reporter-bug.zip](https://github.com/user-attachments/files/16097464/html-reporter-bug.zip)

PR #550, opened 2024-06-04
fix(pwt): ignore private report options from pwt

DudaGod:
## What is done

The static report has stopped working with the new version of the playwright (checked with version 1.44.1). the problem is that playwright has started transferring some of its private fields (`_mode`, `_isTestServer` and `_commandHash` to all connected reports) and this leads to an error:
```
UnknownKeysError: Unknown options: root._mode, root._isTestServer, root._commandHash
```

So I ignore them and everything works fine.

### DIFF
--- a/playwright.ts
+++ b/playwright.ts
@@ -31,7 +31,9 @@ class MyReporter implements Reporter {
     protected _initPromise: Promise<void>;
 
     constructor(opts: Partial<ReporterConfig>) {
-        this._config = parseConfig(_.omit(opts, ['configDir']));
+        const reporterOpts = _.omitBy(opts, (_value, key) => key === 'configDir' || key.startsWith('_'));
+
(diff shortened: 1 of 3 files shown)


playwright.ts
```diff
@@ -31,7 +31,9 @@ class MyReporter implements Reporter {
     protected _initPromise: Promise<void>;
 
     constructor(opts: Partial<ReporterConfig>) {
-        this._config = parseConfig(_.omit(opts, ['configDir']));
+        const reporterOpts = _.omitBy(opts, (_value, key) => key === 'configDir' || key.startsWith('_'));
```
DudaGod, line 34: we don't have unit tests for this, but there are integration tests. That's why I updated the version of pwt to the latest.

sipayRT (approved):

PR #551, opened 2024-06-13
feat: add ability to run cli commands using html-reporter binary

DudaGod:
## What is done

Implement ability to run all cli commands (`gui`, `remove-unused-screens`, `merge-reports`) using html-binary file. Moreover implement `ToolAdapter` structure which allows us to implement support `gui` command with other tools in the near future.

This is the first and largest piece of changes to support `gui` in pwt.


bin/html-reporter
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env node
+'use strict';
+
+(async () => {
+    await require('../build/lib/cli').run();
```
DudaGod, line 5: It is async because I need to import cli commands in runtime and register them

lib/adapters/tool/base.ts
```diff
@@ -0,0 +1,73 @@
+import Testplane, {type Config, type TestCollection} from 'testplane';
```
DudaGod, line 1: I create a separate folder `adapters` so that all adapters are in one place. I'm also going to move `TestResultAdapter` in one of the following PR and create new adapters - `Test`, `Config` (maybe something else)

lib/adapters/tool/base.ts
```diff
@@ -0,0 +1,73 @@
...
+
+interface ToolAdapterOptionsFromCli {
+    toolName: ToolName;
+    configPath?: string;
+}
+
+interface TestplaneAdapterOptionsFromPlugin {
+    toolName: ToolName.Testplane;
+    tool: Testplane;
+    reporterConfig: ReporterConfig;
```
DudaGod, line 20: Used when html-reporter works as plugin (called from testplane binary)

lib/adapters/tool/index.ts
```diff
@@ -0,0 +1,12 @@
+import {TestplaneToolAdapter} from './testplane';
+import {BaseToolAdapter, type BaseToolAdapterOptions} from './base';
+import {ToolName} from '../../constants';
+
+export const makeToolAdapter = (opts: BaseToolAdapterOptions): BaseToolAdapter => {
+    switch (opts.toolName) {
+        case ToolName.Testplane:
+            return TestplaneToolAdapter.create(opts);
+        case ToolName.Playwright:
```
DudaGod, line 9: Pwt is not supported right now

lib/adapters/tool/testplane/index.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    constructor(opts: BaseToolAdapterOptions) {
+        super(opts);
+
+        if ('tool' in opts) {
+            this._tool = opts.tool as TestplaneEnhanced;
+            this._reporterConfig = opts.reporterConfig;
+        } else {
+            // in order to not use static report with gui simultaneously
+            process.env['html_reporter_enabled'] = false.toString();
```
DudaGod, line 39: I need to disable `html-reporter` plugin when it is called from `html-reporter` binary, because in this case I will have gui report and static report in the same time.

lib/adapters/tool/testplane/index.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    get reporterConfig(): ReporterConfig {
+        return this._reporterConfig;
+    }
+
+    get htmlReporter(): HtmlReporter {
+        return this._htmlReporter;
+    }
+
+    initGuiApi(): void {
```
DudaGod, line 64: Inited only when `gui` command is called

lib/adapters/tool/testplane/runner/index.ts
```diff
@@ -1,9 +1,10 @@
 import _ from 'lodash';
```
DudaGod, line 1: All testplane runners moved to testplane folder. They cannot be used for pwt (for example)

lib/cli/commands/gui.js
```diff
@@ -1,17 +1,16 @@
...
-const {Api} = require('../gui/api');
+const {commands} = require('..');
+const runGui = require('../../gui').default;
 
-const {GUI: commandName} = cliCommands;
+const {GUI: commandName} = commands;
 
-module.exports = (program, pluginConfig, testplane) => {
-    // must be executed here because it adds `gui` field in `gemini`, `testplane` and `hermione tool`,
+module.exports = (cliTool, toolAdapter) => {
```
DudaGod, line 8: All cli commands now works with `toolAdapter` instance.

lib/adapters/tool/testplane/index.ts
```diff
@@ -0,0 +1,146 @@
...
+
+        if ('tool' in opts) {
+            this._tool = opts.tool as TestplaneEnhanced;
+            this._reporterConfig = opts.reporterConfig;
+        } else {
+            // in order to not use static report with gui simultaneously
+            process.env['html_reporter_enabled'] = false.toString();
+            this._tool = Testplane.create(opts.configPath) as TestplaneEnhanced;
+
+            const pluginOpts = getPluginOptions(this._tool.config);
```
DudaGod, line 42: Here I need to get options of html-reporter from parsed testplane config.

lib/cli/index.ts
```diff
@@ -0,0 +1,48 @@
...
+    MERGE_REPORTS: 'merge-reports',
+    REMOVE_UNUSED_SCREENS: 'remove-unused-screens'
+} as const;
+
+export const run = async (): Promise<void> => {
+    const program = new Command(pkg.name)
+        .version(pkg.version)
+        .allowUnknownOption()
+        .option('-c, --config <path>', 'path to configuration file')
+        .option('-t, --tool <toolName>', 'tool name which should be run', ToolName.Testplane);
```
DudaGod, line 20: Testplane is used by default

lib/cli/index.ts
```diff
@@ -0,0 +1,48 @@
...
+    for (const commandName of _.values(commands)) {
+        const registerCmd = (await import(path.resolve(__dirname, './commands', commandName))).default;
+
+        registerCmd(program, toolAdapter);
+    }
+
+    program.parse(process.argv);
+};
+
+function preparseOptions(program: Command, options: string[]): Record<string, string | undefined> {
```
DudaGod, line 40: Used from testplane source code.

lib/gui/api/index.ts
```diff
@@ -1,22 +1,19 @@
...
     private _gui: ApiFacade;
 
-    static create<T extends Api>(this: new (testplane: TestplaneWithGui) => T, testplane: TestplaneWithGui): T {
-        return new this(testplane);
+    static create<T extends GuiApi>(this: new () => T): T {
+        return new this();
     }
 
-    constructor(testplane: TestplaneWithGui) {
-        this._gui = testplane.gui = ApiFacade.create();
```
DudaGod, line 19: Tool is not passed here anymore. Because I don't need to set `gui` for other tools (pwt for example). Now it should be specified inside tool adapter instance.

testplane.ts
```diff
@@ -31,9 +31,8 @@ export default (testplane: Testplane, opts: Partial<ReporterOptions>): void => {
         return;
     }
 
-    const htmlReporter = HtmlReporter.create(config, {toolName: ToolName.Testplane});
-
-    (testplane as Testplane & HtmlReporterApi).htmlReporter = htmlReporter;
+    const toolAdapter = makeToolAdapter({toolName: ToolName.Testplane, tool: testplane, reporterConfig: config});
```
DudaGod, line 34: Here I also use tool adapter in order to use the same API inside cli commands

lib/adapters/tool/index.ts
```diff
@@ -0,0 +1,12 @@
+import {TestplaneToolAdapter} from './testplane';
```
KuznetsovRoman, line 1: Lets use dynamic import, and only import testplane tool adapter inside `case ToolName.Testplane`. User, which is using only playwright, will not have `testplane`, which is `./testplane/index.ts` importing

lib/cli/index.ts
```diff
@@ -0,0 +1,48 @@
...
+    const configFileParser = Object.create(program) as Command;
+    configFileParser.options = [].concat(program.options);
+    configFileParser.option('-h, --help');
```
KuznetsovRoman, line 43: What happens on `--help`?

lib/gui/index.ts
```diff
@@ -2,12 +2,10 @@ import type {CommanderStatic} from '@gemini-testing/commander';
 import chalk from 'chalk';
 import opener from 'opener';
 
+import {TestplaneToolAdapter} from '../adapters/tool/testplane';
```
KuznetsovRoman, line 5: I think `import type` should be enough here

KuznetsovRoman (approved):
I dont like that now you have imports from "testplane" in non-testplane-specific files. I think it should be pull out to `/testplane` files, and generic ones should not do this, so when you will try to make playwright adapter, for exaple, you will import "testplane" and get an error. Those should be changed to dynamic imports. Im ok, though, if it would be done in the following PR (its just easier to fix it now)
Other than that, /ok

lib/adapters/tool/base.ts
DudaGod: > I suggest to not import "testplane" in base tool adapter (playwright will be inherited from this class, right?)

I deleted this base class because after implement playwright tool adapter I understand that I only need only an interface. But inside this interface types from tesplane still in use. I will get rid of them in the next PR in which I will add the remaining adapters (TestCollection, Test and Config)

lib/adapters/tool/index.ts
DudaGod: Yeah, you are right. FIxed.

lib/cli/index.ts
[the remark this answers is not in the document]
DudaGod: When I call `npx html-reporter --htlp` I will see all available options and commands. If I don't specify it I will see only options without commands.

DudaGod:
```(bash)
./node_modules/.bin/html-reporter --help

  Usage: html-reporter [options] [command]

  Options:

    -V, --version                       output the version number
    -c, --config <path>                 path to configuration file
    -t, --tool <toolName>               tool name which should be run (default: testplane)
    -h, --help                          output usage information

  Commands:

    gui [options] [paths...]            update the changed screenshots or gather them if they does not exist
    merge-reports [options] [paths...]  merge reports
    remove-unused-screens [options]     remove screenshots which were not used when running tests
```

PR #554, opened 2024-06-19
fix: save RunMode to localStorage

Kabir-Ivan:
Save the state of the run tests button. Default state: run failed.


shadowusr (approved):
need to fix commits history:
- feat -> fix
- remove unnecessary commit
Other than that — LGTM

PR #555, opened 2024-06-19
chore: move "test-adapter" to "adapters/test-result"

DudaGod:
## What is done

- moved the "lib/test-adapter" folder to "lib/adapters/test-result"
- change class names "TestplaneTestAdapter" -> "TestplaneTestResultAdapter" and "TestplaneTestAdapter" -> "PlaywrightTestResultAdapter".

This is done because the implementation of the test adapter will appear in the next PR and there would be a name conflict.


DudaGod:
I merge it, because it looks like there is nothing to review here.

PR #558, opened 2024-06-24
chore: implement adapters: test, testCollection, config

DudaGod:
## What is done

Implement adapters: test, testCollection and config. In order to support gui command for playwright.

Previous PR with implementation tool adapter - https://github.com/gemini-testing/html-reporter/pull/551.
After that PR, I will be able to implement the launch of playwright tests using gui command.

### DIFF
--- a/lib/adapters/test-collection/testplane.ts
+++ b/lib/adapters/test-collection/testplane.ts
@@ -1,0 +1,28 @@
+import {TestplaneTestAdapter} from '../test/testplane';
+import type {TestCollectionAdapter} from './index';
+import type {TestCollection} from 'testplane';
+
+export class TestplaneTestCollectionAdapter implements TestCollectionAdapter {
+    private _testCollection: TestCollection;
+    private _testAdapters: TestplaneTestAdapter[];
+
+    static create<T>(
+        this: new (testCollection: TestCollection) => T,
+        testCollection: TestCollection
+    ): T {
+        return new this(testCollection);
+    }
+
+    constructor(testCollection: TestCollection) {
+        this._testCollection = testCollection;
+        this._testAdapters = this._testCollection.mapTests(test => TestplaneTestAdapter.create(test));
+    }
+
+    get original(): TestCollection {
+        return this._testCollection;
+    }
+
+    get tests(): TestplaneTestAdapter[] {
+        return this._testAdapters;
+    }
+}
--- a/lib/adapters/test/testplane.ts
+++ b/lib/adapters/test/testplane.ts
@@ -1,0 +1,68 @@
+import {TestplaneTestResultAdapter} from '../test-result/testplane';
+import {UNKNOWN_ATTEMPT} from '../../constants';
+
+import type {TestAdapter, CreateTestResultOpts} from './';
+import type {Test, Suite} from 'testplane';
+import type {ReporterTestResult} from '../test-result';
+
+export class TestplaneTestAdapter implements TestAdapter {
+    private _test: Test;
+
+    static create<T extends TestplaneTestAdapter>(this: new (test: Test) => T, test: Test): T {
+        return new this(test);
+    }
+
+    constructor(test: Test) {
+        this._test = test;
+    }
+
+    get original(): Test {
+        return this._test;
+    }
+
+    get id(): string {
+        return this._test.id;
+    }
+
+    get pending(): boolean {
+        return this._test.pending;
+    }
+
+    get disabled(): boolean {
+        return this._test.disabled;
+    }
+
+    get silentlySkipped(): boolean {
+        return isSilentlySkipped(this._test);
+    }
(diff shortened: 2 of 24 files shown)


lib/adapters/test-collection/testplane.ts
```diff
@@ -0,0 +1,26 @@
...
+    constructor(testCollection: TestCollection) {
+        this._originalTestCollection = testCollection;
+    }
+
+    get originalTestCollection(): TestCollection {
+        return this._originalTestCollection;
+    }
+
+    eachTest(cb: TestsCallback): void {
+        this._originalTestCollection.eachTest((test, browserId) => cb(TestplaneTestAdapter.create(test), browserId));
```
DudaGod, line 24: I need to implement this test-collection adapter because I need to send original testCollection in testplane runner (it means I must save it somewhere) and use testAdapter inside gui tool-runner for use equal API (here I create TestAdapter for each test).

lib/adapters/test/testplane.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    static create<T extends TestplaneTestAdapter>(this: new (test: Test) => T, test: Test): T {
+        return new this(test);
+    }
+
+    constructor(test: Test) {
+        this._test = test;
+    }
+
+    get originalTest(): Test {
```
shadowusr, line 19: Is it possible to get rid of this? It leaks implementation details (raw `TestplaneTest` out of abstraction layer `TestAdapter`)

Not critical

lib/adapters/test/testplane.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    get browserId(): string {
+        return this._test.browserId;
+    }
+
+    fullTitle(): string {
+        return this._test.fullTitle();
+    }
+
+    clone(): TestplaneTestAdapter {
```
shadowusr, line 43: Should it really be inside TestAdapter interface and be implemented by all consumers instead of being standalone pure function?

I like to think of these objects as slim, simple and immutable data objects on which we can apply various transformations via pure functions.

For instance, why would you need to implement clone in each adapter if all information you need is exposed via public props and you could just have a single external clone function which could clone any adapter.

Not critical

lib/adapters/test/testplane.ts
```diff
@@ -0,0 +1,54 @@
...
+        return TestplaneTestAdapter.create(this._test.clone());
+    }
+
+    isSilentlySkipped(runnable: Test | Suite = this._test): boolean {
```
shadowusr, line 47: Is it possible to get rid of all methods and keep only `readonly` props? I believe it makes things much more simple and stable.

Not critical

shadowusr (approved):
Overall I think this PR is a huge leap towards making html-reporter tool-agnostic, which is awesome!

All of the comments I wrote are not critical — feel free to merge this without fixing them!
I just wrote things that you may wish to consider, my thoughts.

lib/adapters/test/testplane.ts
DudaGod: I don't like this method either, but I don't find a better solution how to send original test to `getScreenshotPath` from testplane library - https://github.com/gemini-testing/testplane/blob/master/src/config/browser-config.ts#L11.

lib/adapters/test/testplane.ts
DudaGod: > Should it really be inside TestAdapter interface and be implemented by all consumers instead of being standalone pure function?

Runnables inside testplane has `clone` method - https://github.com/gemini-testing/testplane/blob/master/src/test-reader/test-object/test.ts#L23-L30 which must be used to correctly clone instance. For other test adapters I can use default `cloneDeep` from lodash in base class in order to not implement it for each tool.

lib/adapters/test/testplane.ts
DudaGod: > Is it possible to get rid of all methods and keep only readonly props? I believe it makes things much more simple and stable.

I don't understand why methods in class should lead to some problems? in this case, I will have to write the same code inside helpers. And most of these helpers will have matching conditions for the name of the tool and it's much harder to support to my mind.

lib/adapters/test/testplane.ts
```diff
@@ -0,0 +1,54 @@
...
+        return Boolean(runnable.silentSkip || runnable.parent && this.isSilentlySkipped(runnable.parent));
+    }
+
+    formatTestResult(status: TestStatus, attempt: number = UNKNOWN_ATTEMPT): ReporterTestResult {
```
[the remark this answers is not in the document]
DudaGod: Using these methods, I'm trying to get rid of helpers that will have a bunch of if statements with matching on the name of the instrument. Because in this case you will not only implement adapters but also support your tool inside a lot of helpers.

PR #559, opened 2024-06-25
Fix e2e tests

Kabir-Ivan:


package.json
```diff
@@ -20,7 +20,8 @@
     "e2e:build-browsers": "docker build -f test/func/docker/Dockerfile -t html-reporter-browsers:0.0.1 --network host test/func/docker",
     "e2e:build-packages": "npm run --workspace=test/func/packages --if-present build",
     "e2e:generate-fixtures": "npm run --workspace=test/func/fixtures generate",
-    "e2e:launch-browsers": "docker run -it --rm --network=host --add-host=host.docker.internal:0.0.0.0 html-reporter-browsers:0.0.1",
+    "e2e:launch-browsers-local": "if which colima; then docker run -it --rm  --network=host html-reporter-browsers:0.0.1; else docker run -it --rm --network=host --add-host=host.docker.internal:0.0.0.0 html-reporter-browsers:0.0.1; fi",
+    "e2e:launch-browsers": "if which colima; then docker run -it --rm  --network=host yinfra/html-reporter-browsers; else docker run -it --rm --network=host --add-host=host.docker.internal:0.0.0.0 yinfra/html-reporter-browsers; fi",
```
shadowusr, line 24: What's the difference between these two scripts? Only in the yinfra prefix? I'd keep a single script and always use yinfra/... image as I don't see any downsides of this approach.

Also, this may be simplified to something along the lines of:
```
docker run -it --rm --network=host $(which colima >/dev/null || echo --add-host=host.docker.internal:0.0.0.0) yinfra/html-reporter-browsers:0.0.1
```

test/func/docker/Dockerfile
```diff
@@ -1,15 +1,26 @@
-FROM cimg/node:16.20-browsers
+FROM node:18
```
shadowusr, line 1: I see that the resulting images size is around 1GB. Have you tried making it smaller? One obvious thing that comes to mind is that `-slim` version of node:18 image weights only 60MB vs 400MB for the default one: https://hub.docker.com/_/node/tags?page=&page_size=&ordering=&name=18

test/func/docker/Dockerfile
```diff
@@ -1,15 +1,26 @@
...
 
-RUN npm ci && npm run install-chromium
+RUN if [ "$BUILDPLATFORM" = "linux/arm64" ]; then export RELATIVE_PATH="$ARM_CHROMEDRIVER_RELATIVE_PATH"; export URL="$ARM_CHROMEDRIVER_URL"; \
+    else export RELATIVE_PATH="$AMD_CHROMEDRIVER_RELATIVE_PATH"; export URL="$AMD_CHROMEDRIVER_URL"; fi && \
+    wget "$URL" -O "chromedriver.zip" && \
+    unzip "chromedriver" && \
+    if [ "$RELATIVE_PATH" != "" ]; then mv "$RELATIVE_PATH/chromedriver" "/chromedriver"; fi && \
+    apt update -y && \
+    apt install libnss3 -y && \
+    npx playwright install chromium && \
```
shadowusr, line 20: I'm not sure keeping playwright deps/browsers inside the image is a good idea. First, it makes the image much bigger. Secondly, when we'll update playwright in html-reporter, the image won't be updated and this will lead to pretty weird situation. Would it be better to install pwt deps at runtime in workflow, what do you think?

test/func/docker/Dockerfile
```diff
@@ -1,15 +1,26 @@
...
+    npx playwright install-deps && \
+    apt-get install chromium="$CHROMIUM_VERSION" -y
 
-RUN npm install selenium-standalone@9.1.1 -g && \
-    selenium-standalone install --drivers.chrome.version=$CHROME_VERSION
+LABEL com.circleci.preserve-entrypoint=true
 
-ENTRYPOINT selenium-standalone start --drivers.chrome.version=$CHROME_VERSION
+ENTRYPOINT ./chromedriver --port=4444 --whitelisted-ips=""
```
shadowusr, line 26: I remember we tried `--whitelisted-ips` option when testing. Is it actually required?

shadowusr (changes requested):
A few changes are needed. What I expect from working with Docker images in this repo:
- I could pull `yinfra/html-reporter-browsers` from Docker hub and everything will work fine
- I can go to CONTRIBUTING.md and instead of pulling an image from Docker hub follow instructions step-by-step to build an image identical to `yinfra/html-reporter-browsers` and then launch this image locally and run tests using this locally-built image
- I can go to CONTRIBUTING.md and follow step-by-step instruction to publish a new version of `yinfra/...` image.

shadowusr (approved):

PR #570, opened 2024-07-08
chore: move logic with change retries to testplane adapter

DudaGod:
## What is done

- move logic with run tests without retries from `lib/gui/app` to `lib/adapters/tool/testplane`. Because other tools should implement it by themself. 
- add method `runWithoutRetries` which must be implemented in each tool adapter.

### DIFF
--- a/lib/adapters/config/index.ts
+++ b/lib/adapters/config/index.ts
@@ -1,15 +1,9 @@
 import {TestAdapter} from '../test';
 
-export interface BrowserConfigAdapter {
-    readonly id: string;
-    retry: number;
-}
-
 export interface ConfigAdapter {
     readonly tolerance: number;
     readonly antialiasingTolerance: number;
     readonly browserIds: string[];
 
-    getBrowserConfig(browserId: string): BrowserConfigAdapter;
     getScreenshotPath(test: TestAdapter, stateName: string): string;
 }
--- a/lib/adapters/tool/index.ts
+++ b/lib/adapters/tool/index.ts
@@ -31,6 +31,7 @@ export interface ToolAdapter {
     initGuiApi(): void;
     readTests(paths: string[], cliTool: CommanderStatic): Promise<TestCollectionAdapter>;
     run(testCollection: TestCollectionAdapter, tests: TestSpec[], cliTool: CommanderStatic): Promise<boolean>;
+    runWithoutRetries(testCollection: TestCollectionAdapter, tests: TestSpec[], cliTool: CommanderStatic): Promise<boolean>;
 
     updateReference(opts: UpdateReferenceOpts): void;
     handleTestResults(reportBuilder: GuiReportBuilder, eventSource: EventSource): void;
--- a/lib/adapters/tool/testplane/index.ts
+++ b/lib/adapters/tool/testplane/index.ts
@@ -37,15 +37,19 @@ interface OptionsFromPlugin {
     reporterConfig: ReporterConfig;
 }
 
+type RunTestArgs = [TestplaneTestCollectionAdapter, TestSpec[], CommanderStatic];
+
 type Options = ToolAdapterOptionsFromCli | OptionsFromPlugin;
 
 export class TestplaneToolAdapter implements ToolAdapter {
     private _toolName: ToolName;
     private _tool: TestplaneWithHtmlReporter;
-    private _config: ConfigAdapter;
+    private _config: TestplaneConfigAdapter;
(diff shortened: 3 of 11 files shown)


lib/adapters/config/index.ts
```diff
@@ -1,15 +1,9 @@
...
-    readonly id: string;
-    retry: number;
-}
-
 export interface ConfigAdapter {
     readonly tolerance: number;
     readonly antialiasingTolerance: number;
     readonly browserIds: string[];
 
-    getBrowserConfig(browserId: string): BrowserConfigAdapter;
```
DudaGod, line 13: this method is only needed for testplane. Other tools should not implement it.

lib/adapters/tool/index.ts
```diff
@@ -31,6 +31,7 @@ export interface ToolAdapter {
     initGuiApi(): void;
     readTests(paths: string[], cliTool: CommanderStatic): Promise<TestCollectionAdapter>;
     run(testCollection: TestCollectionAdapter, tests: TestSpec[], cliTool: CommanderStatic): Promise<boolean>;
+    runWithoutRetries(testCollection: TestCollectionAdapter, tests: TestSpec[], cliTool: CommanderStatic): Promise<boolean>;
```
DudaGod, line 34: Add new method in order to not forget in new adapters to implement logic with disable retries before run tests

lib/adapters/tool/testplane/index.ts
```diff
@@ -37,15 +37,19 @@ interface OptionsFromPlugin {
...
 export class TestplaneToolAdapter implements ToolAdapter {
     private _toolName: ToolName;
     private _tool: TestplaneWithHtmlReporter;
-    private _config: ConfigAdapter;
+    private _config: TestplaneConfigAdapter;
     private _reporterConfig: ReporterConfig;
     private _htmlReporter: HtmlReporter;
     private _guiApi?: GuiApi;
+    private _browserConfigs: ReturnType<Config['forBrowser']>[];
+    private _retryCache: Record<string, number>;
```
DudaGod, line 52: This logic just moved from `lib/gui/app`

test/unit/lib/adapters/tool/testplane/index.ts
```diff
@@ -340,6 +340,113 @@ describe('lib/adapters/tool/testplane/index', () => {
         });
     });
 
+    describe('runWithoutRetries', () => {
```
DudaGod, line 343: Use tests from `test/lib/gui/app`. Here I just brought them to the necessary form

lib/gui/tool-runner/index.ts
```diff
@@ -321,7 +321,9 @@ export class ToolRunner {
     async run(tests: TestSpec[] = []): Promise<boolean> {
         const testCollection = this._ensureTestCollection();
 
-        return this._toolAdapter.run(testCollection, tests, this._globalOpts);
+        return _.isEmpty(tests)
```
shadowusr, line 324: Hmm it's not entirely clear to me why if tests are empty here we use regular run and runWithoutRetries otherwise. We could save _.isEmpty into a boolean variable with descriptive name or leave a comment. Although I may be just missing something.

lib/adapters/tool/testplane/index.ts
```diff
@@ -119,6 +126,13 @@ export class TestplaneToolAdapter implements ToolAdapter {
         return runner.run((collection) => this._tool.run(collection, {grep, sets, browsers, devtools, replMode}));
     }
 
+    async runWithoutRetries(...args: RunTestArgs): Promise<boolean> {
```
shadowusr, line 129: I see that this method was just moved from the other file, but when do we need it, what's the use-case?

shadowusr (approved):
Looks good to me 👍 
Left a few (non-critical) questions

lib/adapters/tool/testplane/index.ts
DudaGod: It was moved from `lib/gui/app`, which is a common file for all adapters. But for other tools the change in the number of retries can be performed in completely different ways. For example for playwright I should specify option `--retries 0` in cli command, like this `playwright test --retries 0`.

lib/gui/tool-runner/index.ts
DudaGod: In case when user click on `run` command in UI interface we send empty array, which means that we must run all tests with all available retires. But when user want to retry only one test or suite we disable retries.

I will add boolean variable with descriptive name.

PR #571, opened 2024-07-09
feat: support playwright in gui mode

DudaGod:
## What is done

- write adapters for playwright tool in order to support gui mode


lib/adapters/config/playwright.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export type PwtConfig = FullConfig & {
+    testDir?: string;
+    snapshotDir?: string;
+    snapshotPathTemplate?: string;
+    projects?: PwtProject[]
+}
+
+export const DEFAULT_BROWSER_ID = 'chromium';
+const DEFAULT_SNAPSHOT_PATH_TEMPLATE = '{snapshotDir}/{testFileDir}/{testFileName}-snapshots/{arg}{-projectName}{-snapshotSuffix}{ext}';
```
DudaGod, line 21: This default value I get from pwt source code. And it is not exported outside.

lib/adapters/config/playwright.ts
```diff
@@ -0,0 +1,114 @@
...
+    private _config: PwtConfig;
+    private _browserIds: string[];
+
+    static create<T extends PlaywrightConfigAdapter>(this: new (config: PwtConfig) => T, config: PwtConfig): T {
+        return new this(config);
+    }
+
+    constructor(config: PwtConfig) {
+        this._config = config;
+        this._browserIds = _.isEmpty(this._config.projects) ? [DEFAULT_BROWSER_ID] : this._config.projects.map(prj => prj.name).filter(Boolean);
```
DudaGod, line 33: if the user does not specify the browser in the `projects` section then `chromium` will be used by default. So I support this behaviour.

lib/adapters/config/playwright.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    get original(): PwtConfig {
+        return this._config;
+    }
+
+    get tolerance(): number {
+        return 2.3;
```
DudaGod, line 41: Default value in order to use in looks-same (when click on `Find same diffs` in UI)

lib/adapters/config/playwright.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    get antialiasingTolerance(): number {
+        return 4;
```
DudaGod, line 45: the same as above

lib/adapters/config/playwright.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    get browserIds(): string[] {
+        return this._browserIds;
+    }
+
+    getScreenshotPath(test: TestAdapter, stateName: string): string {
```
DudaGod, line 52: here is a complicated logic that I used from pwt (with minor changes from me to correctly support it) - https://github.com/microsoft/playwright/blob/main/packages/playwright/src/worker/testInfo.ts#L452-L473

lib/adapters/config/playwright.ts
```diff
@@ -0,0 +1,114 @@
...
+            .replace(/\{(.)?testFileName\}/g, '$1' + parsedRelativeTestFilePath.base)
+            .replace(/\{(.)?testFilePath\}/g, '$1' + test.file)
+            .replace(/\{(.)?arg\}/g, '$1' + path.join(parsedSubPath.dir, parsedSubPath.name))
+            .replace(/\{(.)?ext\}/g, parsedSubPath.ext ? '$1' + parsedSubPath.ext : '');
+
+        return path.normalize(path.resolve(snapshotPath));
+    }
+}
+
+function sanitizeForFilePath(s: string): string {
```
DudaGod, line 85: these helpers were also used from pwt

lib/adapters/test-result/playwright.ts
```diff
@@ -6,20 +6,27 @@ import stripAnsi from 'strip-ansi';
...
     ImageInfoDiff,
-    ImageInfoFull, ImageInfoNoRef, ImageInfoPageError, ImageInfoPageSuccess, ImageInfoSuccess,
+    ImageInfoFull, ImageInfoNoRef, ImageInfoPageError, ImageInfoPageSuccess, ImageInfoSuccess, ImageInfoUpdated,
     ImageSize,
     TestError
 } from '../../types';
 import type {CoordBounds} from 'looks-same';
 
-export type PlaywrightAttachment = PlaywrightTestResult['attachments'][number];
+export type PlaywrightAttachment = PlaywrightTestResult['attachments'][number] & {
```
DudaGod, line 22: Here I support values when image is updated (got it from testplane)

lib/adapters/test-result/playwright.ts
```diff
@@ -42,13 +49,25 @@ export enum ImageTitleEnding {
     Previous = '-previous.png'
 }
 
+export interface TestResultWithGuiStatus extends Omit<PlaywrightTestResult, 'status'> {
+    status: PlaywrightTestResult['status'] | TestStatus.RUNNING | TestStatus.UPDATED;
```
DudaGod, line 53: I need two new statuses: running and updated. I just get it from `TestStatus`

lib/adapters/test-result/playwright.ts
```diff
@@ -127,7 +146,8 @@ const getImageData = (attachment: PlaywrightAttachment | undefined): ImageFile |
 
     return {
         path: attachment.path as string,
-        size: _.pick(sizeOf(attachment.path as string), ['height', 'width']) as ImageSize
+        size: !attachment.size ? _.pick(sizeOf(attachment.path as string), ['height', 'width']) as ImageSize : attachment.size,
```
DudaGod, line 149: When image is updated I already have size value and don't want to recalculate it again

lib/adapters/test-result/playwright.ts
```diff
@@ -230,6 +259,14 @@ export class PlaywrightTestResultAdapter implements ReporterTestResult {
                     error: _.pick(error, ['message', 'name', 'stack']),
                     actualImg
                 } satisfies ImageInfoNoRef;
+            } else if (expectedAttachment?.isUpdated && expectedImg && actualImg) {
```
DudaGod, line 262: got it from testplane source code

lib/adapters/test/playwright.ts
```diff
@@ -0,0 +1,137 @@
...
+                file: this._test.file
+            },
+            parent: {
+                project: () => ({
+                    name: this._test.browserName
+                })
+            }
+        } as unknown as TestCase;
+
+        const attachments = assertViewResults.map(assertViewResult => {
```
DudaGod, line 82: Here I need to correctly handle `assertViewResults` and convert it to attachments

lib/adapters/test/playwright.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    get titlePath(): string[] {
+        return this._test.titlePath;
+    }
+
+    createTestResult(opts: CreateTestResultOpts): ReporterTestResult {
+        const {status, attempt = UNKNOWN_ATTEMPT, assertViewResults = []} = opts;
+
+        const testCase = {
```
DudaGod, line 68: Here I need to generate `testCase` from raw result which I get from client when user click on `accept` button

lib/adapters/tool/playwright/index.ts
```diff
@@ -0,0 +1,311 @@
...
+        const projectArgs = Array.from(browserNames).flatMap(broName => [haveProjectsInConfig ? '--project' : '--browser', broName]);
+        args.push(...projectArgs);
+    } else if (!haveProjectsInConfig) {
+        args.push(...['--browser', DEFAULT_BROWSER_ID]);
+    }
+
+    return args;
+}
+
+function getPluginOptions(config: PlaywrightConfigAdapter['original']): Partial<ReporterConfig> {
```
DudaGod, line 291: Here I get options from config file inside our reporter - `html-reporter/playwright`

lib/adapters/tool/playwright/index.ts
```diff
@@ -0,0 +1,311 @@
...
+        process.exit(1);
+    }
+}
+
+function readPwtConfig(opts: ToolAdapterOptionsFromCli): {configPath: string, config: FullConfig} {
+    const configPaths = opts.configPath ? [opts.configPath] : DEFAULT_CONFIG_PATHS;
+    let originalConfig!: FullConfig;
+    let resolvedConfigPath!: string;
+
+    const revertTransformHook = setupTransformHook();
```
DudaGod, line 215: Here I need to use pirates with babel in order to correctly transpile config on `.ts`, `.mts` and `.mjs`. Other extensions are skipped

lib/adapters/tool/playwright/index.ts
```diff
@@ -0,0 +1,311 @@
...
+
+    for (const configPath of configPaths) {
```
DudaGod, line 217: Here I try to find config file in project if user not specify it by himself

lib/adapters/tool/playwright/index.ts
```diff
@@ -0,0 +1,311 @@
...
+import type {PwtEventMessage} from './reporter';
+
+export const DEFAULT_CONFIG_PATHS = [
+    `${ToolName.Playwright}.config.ts`,
+    `${ToolName.Playwright}.config.js`,
+    `${ToolName.Playwright}.config.mts`,
+    `${ToolName.Playwright}.config.mjs`,
+    `${ToolName.Playwright}.config.cts`,
+    `${ToolName.Playwright}.config.cjs`
+];
```
DudaGod, line 39: Got it from source code of pwt

lib/adapters/tool/playwright/index.ts
```diff
@@ -0,0 +1,311 @@
...
+    initGuiApi(): void {
+        this._guiApi = GuiApi.create();
+    }
+
+    async readTests(): Promise<PlaywrightTestCollectionAdapter> {
+        const stdout = await new Promise<string>((resolve, reject) => {
+            // specify default browser in order to get correct stdout with browser name
+            const browserArgs = this._haveProjectsInConfig ? [] : ['--browser', DEFAULT_BROWSER_ID];
+
+            const child = spawn(this._pwtBinaryPath, ['test', '--list', '--reporter', 'list', '--config', this._configPath, ...browserArgs]);
```
DudaGod, line 109: Here I use `--list` option in order to get all tests info (browser, file and title)

lib/adapters/tool/playwright/index.ts
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    initGuiApi(): void {
+        this._guiApi = GuiApi.create();
+    }
+
+    async readTests(): Promise<PlaywrightTestCollectionAdapter> {
+        const stdout = await new Promise<string>((resolve, reject) => {
+            // specify default browser in order to get correct stdout with browser name
+            const browserArgs = this._haveProjectsInConfig ? [] : ['--browser', DEFAULT_BROWSER_ID];
```
DudaGod, line 107: I need to specify `chromium` if user not use `projects` in order to get correct stdout with browser name

lib/adapters/tool/playwright/index.ts
```diff
@@ -0,0 +1,311 @@
...
+                    return reject(new Error(`Playwright process with reading tests exited with code: ${code}, stderr: ${stderr}`));
+                }
+
+                resolve(stdout);
+            });
+        });
+
+        const stdoutByLine = stdout.split('\n').map(v => v.trim());
+        const startIndex = stdoutByLine.findIndex(v => v === 'Listing tests:');
+        const endIndex = stdoutByLine.findIndex(v => /total: \d+ tests? in \d+ file/i.test(v));
```
DudaGod, line 129: Stout of this command looks like this:
```
Listing tests:
  example.spec.js:4:5 › suite #js › has title #1
  example.spec.js:16:5 › suite #js › has title #2
Total: 2 tests in 1 file
```

So I need to get all info between `Listing tests` and `Total: ...`

lib/adapters/tool/playwright/index.ts
```diff
@@ -0,0 +1,311 @@
...
+        if (!this._reportBuilder || !this._eventSource) {
+            throw new Error('"reportBuilder" and "eventSource" instances must be initialize before run tests');
+        }
+
+        const queue = new PQueue({concurrency: os.cpus().length});
+
+        return new Promise((resolve, reject) => {
+            const args = ([] as string[]).concat(prepareRunArgs(tests, this._configPath, this._haveProjectsInConfig), runArgs);
+            const child = spawn(this._pwtBinaryPath, args, {
+                stdio: ['inherit', 'inherit', 'inherit', 'ipc']
```
DudaGod, line 163: Use ipc in order to communicate between special reporter which is started inside pwt and html-reporter process.

lib/adapters/tool/playwright/index.ts
```diff
@@ -0,0 +1,311 @@
...
+function prepareRunArgs(tests: TestSpec[], configPath: string, haveProjectsInConfig: boolean): string[] {
+    const testNames = new Set<string>();
+    const browserNames = new Set<string>();
+
+    for (const {testName, browserName} of tests) {
+        testNames.add(testName);
+        browserNames.add(browserName);
+    }
+
+    const args = ['test', '--reporter', path.resolve(__dirname, './reporter'), '--config', configPath];
```
DudaGod, line 275: I use my custom reporter to handle results from pwt.

lib/adapters/tool/playwright/reporter.ts
```diff
@@ -0,0 +1,36 @@
...
+import type {TestResultWithGuiStatus} from '../../test-result/playwright';
+
+export type PwtEventMessage = {
+    test: TestCase;
+    result: TestResultWithGuiStatus;
+    browserName: string;
+    titlePath: string[];
+};
+
+export default class MyReporter implements Reporter {
```
DudaGod, line 15: Reporter which is used to get info about test results

lib/adapters/tool/playwright/reporter.ts
```diff
@@ -0,0 +1,36 @@
+import stringify from 'json-stringify-safe';
```
DudaGod, line 1: Data from pwt has circular links which I can't pass between processes

test/unit/lib/adapters/config/playwright.ts
```diff
@@ -0,0 +1,208 @@
+import path from 'node:path';
+import sinon from 'sinon';
+
+import {PlaywrightConfigAdapter, DEFAULT_BROWSER_ID, type PwtConfig, type PwtProject} from '../../../../../lib/adapters/config/playwright';
```
shadowusr, line 4: Those `../../..` are ugly as hell =( Would be nice to import them via `@site/lib/adapters`. Absolutely not suggesting to implement it in this PR, just thinking out loud, maybe we need an issue for this...

lib/cli/commands/remove-unused-screens/index.js
```diff
@@ -39,6 +40,10 @@ module.exports = (program, toolAdapter) => {
         .option('--skip-questions', 'do not ask questions during execution (default values will be used)')
         .on('--help', () => logger.log(getHelpMessage(toolName)))
         .action(async (options) => {
+            if (toolName !== ToolName.Testplane) {
+                throw new Error(`CLI command "${commandName}" supports only for "${ToolName.Testplane}" tool`);
```
shadowusr, line 44: typo: `CLI command X supports only Y tool`

lib/cli/commands/merge-reports.js
```diff
@@ -3,17 +3,24 @@
...
+
     program
         .command(`${commandName} [paths...]`)
         .allowUnknownOption()
         .description('merge reports')
         .option('-d, --destination <destination>', 'path to directory with merged report', toolAdapter.reporterConfig.path)
         .option('-h, --header <header>', 'http header for databaseUrls.json files from source paths', collect, [])
         .action(async (paths, options) => {
+            if (toolName !== ToolName.Testplane) {
+                throw new Error(`CLI command "${commandName}" supports only for "${ToolName.Testplane}" tool`);
```
shadowusr, line 21: The same typo here.

lib/adapters/tool/playwright/index.ts
shadowusr: I am 100% sure you've checked this, but do they really lack json-reporter so that we didn't need to parse it like this?

lib/adapters/tool/playwright/index.ts
shadowusr: Surprised pwt can't just return the parsed config via some API or at least CLI.

lib/adapters/tool/playwright/index.ts
```diff
@@ -0,0 +1,311 @@
...
+
+            child.on('error', (data) => reject(data));
+
+            child.on('exit', (code) => {
+                queue.onIdle().then(() => resolve(!code));
+            });
+        });
+    }
+
+    // Can't handle test results here because pwt does not provide api for this, so save instances and use them in custom reporter
```
shadowusr, line 196: By saying "in custom reporter" do you mean those ipc handlers above? (just making sure I got it right)

lib/adapters/tool/playwright/index.ts
```diff
@@ -0,0 +1,311 @@
...
+
+    for (const {testName, browserName} of tests) {
+        testNames.add(testName);
+        browserNames.add(browserName);
+    }
+
+    const args = ['test', '--reporter', path.resolve(__dirname, './reporter'), '--config', configPath];
+
+    if (testNames.size > 0) {
+        args.push('--grep', Array.from(testNames).join('|'));
```
shadowusr, line 278: Will this work fine if test name contains brackets, etc or do we need additional escaping?

lib/adapters/tool/playwright/index.ts
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    if (browserNames.size > 0) {
+        const projectArgs = Array.from(browserNames).flatMap(broName => [haveProjectsInConfig ? '--project' : '--browser', broName]);
```
shadowusr, line 282: The boolean flag should be named `has...` (not critical, but as far as I know the convention is that.)

shadowusr, line 282: Hmm what's the difference between `--browser` and `--project`? I see only `--project` in pwt docs.

lib/adapters/tool/playwright/ipc.ts
```diff
@@ -0,0 +1,17 @@
+import _ from 'lodash';
+import type {ChildProcess} from 'node:child_process';
+
+export default {
+    emit: (event: string, data: Record<string, unknown> = {}): void => {
+        process.send && process.send({event, ...data});
+    },
+    on: <T extends Record<string, unknown>>(event: string, handler: (msg: T & {event: string}) => void, proc: ChildProcess | NodeJS.Process = process): void => {
```
shadowusr, line 8: Hmm didn't we have more convenient tools for IPC somewhere? So that we didn't need to implement this from scratch. Not critical.

lib/adapters/config/playwright.ts
shadowusr: Idk, is it worth leaving this comment in the code, what do you think?

shadowusr (approved):
Looks great to me! Can't imagine what you had to go through to make this work, respect =)

:warning: I see that E2E workflow is skipped once again in this PR. Please run it locally.

*Ideally*, it would be really nice and I think logical to turn on the same GUI tests that we have for testplane for playwright also: https://github.com/gemini-testing/html-reporter/tree/master/test/func/tests/common-gui

lib/adapters/tool/playwright/ipc.ts
DudaGod: I used a similar solution from testplane - https://github.com/gemini-testing/testplane/blob/8c82cd4468eddd7a2a7dafac119fd4f840fdacd4/src/utils/ipc.js. There are only ~10 lines of simple code here I don't think that we should find some library for this (but I searched on the Internet and didn't find simple solution)

test/unit/lib/adapters/config/playwright.ts
DudaGod: > Those ../../.. are ugly as hell =(

Yes, I suffer from this when I write tests. We can write path like - `'lib/adapters/config/playwright'`due to the use of `app-module-path`, but he loses types on the way so I don't use it.

In general, it seems that we just need to move to jest and put the tests near to the code, then there will be no problems.

> Would be nice to import them via @site/lib/adapters

I didn't understand how it would work? What the `@site` means?

PR #573, opened 2024-07-12
refactor: refactor test body using GravityUI

Kabir-Ivan:


shadowusr:
This panel in GUI mode has a few issues:


- icons are not aligned
- icons should be from gravity as well: https://gravity-ui.com/icons — they have everything we need

shadowusr:
Entries in history are not selectable. We should be able to select and copy contents of history as a regular text.

shadowusr:
Eye button in the "switch accept mode" is not positioned right:

shadowusr:
Diff mode select in switch accept mode isn't displaying full option name, despite having more than enough space to do so:

shadowusr:
Here we should have share button, not copy button:

shadowusr:
Eye button appears disabled, even though URL is available and upon click it actually sends us to the corresponding page. It shouldn't be disabled then.

shadowusr:
There's no difference between currently running test and never launched test:


I suggest we draw a rotating spinner instead of the run button for currently running tests: https://gravity-ui.com/components/uikit/spin

shadowusr:
"retry" buttons do not become disabled while tests are running, but they should:

shadowusr:
Retry buttons have arrows pointed in different directions and of different sizes. I suggest making them of the exact same size and pointed to the right.

shadowusr:
Tests that were never run and currently running tests have a weird collapsible arrow that does nothing. It either shouldn't be there at all or should display something.

shadowusr:
Slider on the onion skin tab is not from Gravity UI, but it should be:


https://gravity-ui.com/components/uikit/slider

shadowusr:
Arrows on buttons in switch accept mode are not centered:

shadowusr:
Undo should have an icon from Gravity UI too:


The one you're using in the retry button next to test title may be a good fit.

shadowusr:
When I click on/off checkboxes, test titles stutter, but they shouldn't:

lib/static/components/section/title/browser.jsx
```diff
@@ -22,7 +22,7 @@ const BrowserTitle = (props) => {
             viewModes: ViewMode.ALL,
             expand: EXPAND_ALL
         });
-    };
+    }, [window.location.href, props.browserName, props.testName, props.retryIndex]);
```
shadowusr, line 25: `window.location.href` is not reactive as far as i know. So placing it in dependencies will do nothing.

shadowusr (changes requested):
Formatting is pretty inconsistent. I understand that linter doesn't work now, but let's at least keep formatting right in the new code. What I noticed: `{item}` vs `{ item }`, incorrect indents

Also I'm still in doubts about the new colors for attempts. I do like them, but I'll ask our users if they like them too. Because some can definitely say that they are too bleak and red is indistinguishable from pink.

Other than that — great PR and good job! Just need a few more fixes.

shadowusr:
And one more thing: when launching a test in GUI, attempt which is running appears yellow. But should be white, because yellow in attempt switcher is reserved for future use

shadowusr (approved):
Stacktraces need to be in mono font. Other than that, everything's great!

PR #574, opened 2024-07-17
feat: add "timestamp" field to redux storage

DudaGod:
## What is done:

- add "timestamp" field to redux storage in order to use from plugins
- filed "date" will be removed in next major version

### DIFF
--- a/lib/static/modules/reducers/timestamp.js
+++ b/lib/static/modules/reducers/timestamp.js
@@ -1,14 +1,12 @@
 import actionNames from '../action-names';
-import {dateToLocaleString} from '../utils';
 
-// TODO: remove in next major (should use timestamp instead)
 export default (state, action) => {
     switch (action.type) {
         case actionNames.INIT_GUI_REPORT:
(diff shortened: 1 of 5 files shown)


lib/static/modules/reducers/timestamp.js
```diff
@@ -0,0 +1,17 @@
+import actionNames from '../action-names';
+
+export default (state, action) => {
+    switch (action.type) {
+        case actionNames.INIT_GUI_REPORT:
+        case actionNames.INIT_STATIC_REPORT: {
+            const {timestamp} = action.payload;
+
+            console.log('timestamp:', timestamp);
```
KuznetsovRoman, line 9: extra console log

KuznetsovRoman (approved):
Let's also stop using "date" in this PR ourselves.
I suggest to change this label so it would use `timestamp` instead of `date`: https://github.com/gemini-testing/html-reporter/blob/master/lib/static/components/controls/report-info.jsx#L18
Otherwise, /ok

PR #575, opened 2024-07-17
fix: return correct test result status

KuznetsovRoman:
Now (test has failed status, and no error message is shown, though there was an error):
image

After this PR (test has error status and error message is shown):
image

### DIFF
--- a/lib/adapters/test-result/testplane.ts
+++ b/lib/adapters/test-result/testplane.ts
@@ -6,7 +6,7 @@ import {ValueOf} from 'type-fest';
 
 import {getCommandsHistory} from '../../history-utils';
 import {ERROR, FAIL, SUCCESS, TestStatus, UNKNOWN_SESSION_ID, UPDATED} from '../../constants';
-import {getError, hasFailedImages, isImageDiffError, isNoRefImageError, wrapLinkByTag} from '../../common-utils';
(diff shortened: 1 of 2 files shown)


lib/adapters/test-result/testplane.ts
```diff
@@ -32,7 +32,7 @@ export const getStatus = (eventName: ValueOf<Testplane['events']>, events: Testp
     } else if (eventName === events.TEST_PENDING) {
         return TestStatus.SKIPPED;
     } else if (eventName === events.RETRY || eventName === events.TEST_FAIL) {
-        return hasFailedImages(testResult.assertViewResults as ImageInfoFull[]) ? TestStatus.FAIL : TestStatus.ERROR;
+        return isAssertViewError(testResult.err) ? TestStatus.FAIL : TestStatus.ERROR;
```
KuznetsovRoman, line 35: Test result with errors could still has failed images

DudaGod: Why you check here only assert view error and don't check diff (`isImageDiffError`) or no ref image (`isNoRefImageError`)?

DudaGod (approved):
otherwise /ok

lib/adapters/test-result/testplane.ts
KuznetsovRoman: Because testResult can only have assert view error.
`isImageDiffError` and `isNoRefImageError` - are image errors, and not the result ones. So if testResult has an image with `isImageDiffError` or `isNoRefImageError` - it would have `AssertViewError`

PR #576, opened 2024-07-18
fix: support hermione

DudaGod:
## What is done

- fix support hermione, it was broken in #551
- specify pwt in peer deps

### DIFF
--- a/lib/adapters/tool/testplane/index.ts
+++ b/lib/adapters/tool/testplane/index.ts
@@ -1,5 +1,6 @@
 import _ from 'lodash';
-import Testplane, {type Config} from 'testplane';
+import type Testplane from 'testplane';
+import type {Config} from 'testplane';
 import type {CommanderStatic} from '@gemini-testing/commander';
 
 import {TestplaneTestCollectionAdapter} from '../../test-collection/testplane';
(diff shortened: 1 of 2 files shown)


lib/adapters/tool/testplane/index.ts
```diff
@@ -1,5 +1,6 @@
 import _ from 'lodash';
-import Testplane, {type Config} from 'testplane';
+import type Testplane from 'testplane';
```
DudaGod, line 2: Testplane is not imported here, because this file is used also for hermione tool

lib/adapters/tool/testplane/index.ts
```diff
@@ -194,6 +197,31 @@ function getPluginOptions(config: Config): Partial<ReporterConfig> {
     return defaultOpts;
 }
 
+function createTool(configPath?: string): TestplaneWithHtmlReporter {
+    let tool!: TestplaneWithHtmlReporter;
+
+    for (const toolName of SUPPORTED_TOOLS) {
+        try {
+            // eslint-disable-next-line @typescript-eslint/no-var-requires
+            const Tool = require(toolName).default;
```
DudaGod, line 206: I use `require` because in other way i should use `await import` in constructor (can't be async) or in static method `create`. But it is called inside plugin entry point, which is not async either.

lib/adapters/tool/testplane/index.ts
```diff
@@ -1,5 +1,6 @@
 import _ from 'lodash';
-import Testplane, {type Config} from 'testplane';
+import type Testplane from 'testplane';
+import {type Config} from 'testplane';
```
KuznetsovRoman, line 3: Why not "`import type {Config} from 'testplane'`"?

KuznetsovRoman (approved):

PR #581, opened 2024-07-24
fix(pwt): correctly import ts config with types

DudaGod:
## What is done

- move "@gemini-testing/commander" from devDeps to deps
- correctly import ts config with types for playwright


DudaGod:
Looks like there is nothing to review. So I merge it.

PR #582, opened 2024-07-28
chore: migrate to react 18

shadowusr:
What's done:
- `React 16` -> `React 18`
- `Enzyme` -> `@testing-library/react`
- Updated a bunch of other packages
- Tests are mostly updated to testing-library without major changes, but some tests were refactored here and there. Notably, `test/unit/lib/static/components/modals/screenshot-accepter/index.jsx` was rewritten from ground up.
- `rc-resize-observer` -> `@react-hook/resize-observer`
- removed `styled-components` that were used in just one place

⚠️ You may notice that some tests became less "unit'y" and lean towards integration tests. Here's rationale:
- [You shouldn't mock selectors or store at all](https://redux.js.org/usage/writing-tests#guiding-principles)
- [The app should be tested the way users interact with it](https://testing-library.com/docs/guiding-principles)
- [You shouldn't mock components](https://testing-library.com/docs/react-testing-library/faq)

⚠️ Also, some tests have a lot of duplicate setup code. I've spent a while trying to make this better, but in the end it comes to a choice whether a test is self-contained&easily readable or there's no duplication. I chose readability. If you see a concrete way of achieving both, I'd be glad to hear!


lib/static/components/controls/menu-bar.jsx
```diff
@@ -33,7 +33,7 @@ class MenuBar extends Component {
         }
 
         return (
-            <div className="menu-bar" data-test-id='menu-bar'>
+            <div className="menu-bar" data-qa='menu-bar'>
```
shadowusr, line 36: Gravity UI uses `data-qa` attribute and in some places it's not configurable. So I decided to use `data-qa` everywhere in the app for consistency.

lib/static/components/header/summary/dbBtn.jsx
```diff
@@ -11,16 +10,14 @@ const ForwardedDbBtn = React.forwardRef(function DbBtn({fetchDbDetails}, ref) {
...
-            <Button
-                view={isFailed ? 'flat-danger' : 'flat'}
-            >
-                <div className='db-info'>
-                    <ChevronDown/>
-                    {content}
-                </div>
-            </Button>
-        </Ref>
+        <Button
```
shadowusr, line 13: `<Ref>` is deprecated.

lib/static/components/state/screenshot/with-synced-scale.jsx
```diff
@@ -1,116 +1,116 @@
-import React, {Component} from 'react';
+import useResizeObserver from '@react-hook/resize-observer';
```
shadowusr, line 1: Got rid of `rc-resize-observer` in favor of more modern hook-based solution.

rc-resize-observer uses obsolete API and it's unclear when it will be fixed: https://github.com/react-component/resize-observer/issues/208

lib/static/components/state/state-error.jsx
```diff
@@ -5,7 +5,7 @@ import {connect} from 'react-redux';
 import {bindActionCreators} from 'redux';
 import PropTypes from 'prop-types';
 import {isEmpty, map, isFunction} from 'lodash';
-import ReactHtmlParser from 'react-html-parser';
+import ReactHtmlParser from 'html-react-parser';
```
shadowusr, line 8: We used two different packages that do one thing =)

.mocharc.js
```diff
@@ -3,5 +3,5 @@
 module.exports = {
     extension: ["js", "jsx", "ts", "tsx"],
     recursive: true,
-    require: ["./test/ts-node", "./test/setup", "./test/assert-ext", "jsdom-global/register"],
+    require: ["./test/ts-node", "./test/setup", "./test/assert-ext", "./test/configure-testing-library"],
```
KuznetsovRoman, line 6: I would put all of these setup files in extra folder: "test/setup/*", for example.

lib/static/components/header/summary/dbBtn.jsx
KuznetsovRoman: i recall it was usefull when it was wrapping semantic-ui-react component, as they can't pass `ref` the usual way. Useless with gravity-ui, though

lib/static/components/modals/screenshot-accepter/index.jsx
```diff
@@ -208,13 +208,13 @@ class ScreenshotAccepter extends Component {
         return imageId;
     }
 
-    _preloadAdjacentImages(offset = PRELOAD_IMAGE_COUNT) {
-        const screensCount = size(this.state.stateNameImageIds);
-        const previosImagesIndex = (screensCount + this.state.activeImageIndex - offset) % screensCount;
-        const nextImagesIndex = (this.state.activeImageIndex + offset) % screensCount;
+    _preloadAdjacentImages(activeImageIndex, stateNameImageIds, offset = PRELOAD_IMAGE_COUNT) {
```
KuznetsovRoman, line 211: Why does it need to be changed?

lib/static/components/state/state-error.jsx
KuznetsovRoman: It is called diversification)

package.json
```diff
@@ -105,6 +105,7 @@
     "json-stringify-safe": "^5.0.1",
     "lodash": "^4.17.4",
     "looks-same": "^8.2.1",
+    "msw": "^2.3.4",
```
KuznetsovRoman, line 108: Why isn't it in devDeps?

package.json
```diff
@@ -185,43 +187,40 @@
...
+    "reapop": "^4.2.2",
     "reduce-reducers": "^1.0.4",
     "redux": "^4.0.5",
     "redux-logger": "^3.0.6",
     "redux-mock-store": "^1.5.1",
     "redux-thunk": "^2.2.0",
     "reselect": "^4.1.6",
     "rimraf": "^2.6.3",
-    "semantic-ui-react": "^1.2.1",
+    "semantic-ui-react": "^3.0.0-beta.2",
```
KuznetsovRoman, line 219: Why `beta`?
Other versions does not support react@18?

test/func/tests/plugins/tests-menu-bar-plugin.testplane.js
```diff
@@ -12,6 +12,9 @@ describe('Test menu bar plugin', function() {
         await browser.$(selector).waitForDisplayed();
         await browser.$(selector).click();
         await browser.$(menuSelector).waitForDisplayed();
+
+        // Pause prevents flaky screenshots due to dropdown shadow rendering. No, screenshot delay doesn't help.
+        await browser.pause(1000);
```
KuznetsovRoman, line 17: Does it need to be that high?

test/unit/lib/static/components/bullet.jsx
```diff
@@ -18,51 +20,28 @@ describe('<Bullet />', () => {
     it('should render simple bullet if checkboxes are disabled', () => {
         useLocalStorageStub.withArgs('showCheckboxes', false).returns([false]);
 
-        const component = mount(<Bullet bulletClassName='bullet_type-simple' />);
+        const component = render(<Bullet bulletClassName='bullet_type-simple' />);
 
-        assert.isTrue(component.find('.bullet_type-simple').exists());
+        expect(component.getByTestId('bullet-icon')).to.exist;
```
KuznetsovRoman, line 25: `.to.exist` without `()` at the end feels super weird. Why would `chai` do this

test/unit/lib/static/components/bottom-progress-bar/index.jsx
```diff
@@ -30,7 +30,7 @@ describe('<BottomProgressBar/> component', () => {
                 visibleRootSuiteIds: ['id1', 'id2', 'id3', 'id4', 'id5']
             });
 
-            assert.equal(component.find('.bottom-progress-bar__progress-bar').prop('style').width, '0%');
+            assert.equal(component.container.querySelector('.bottom-progress-bar__progress-bar').style.width, '0%');
```
KuznetsovRoman, line 33: Why do we need to access `container` property? Can't we call `component.getByClassName` or something like that?

test/unit/lib/static/components/controls/browser-list/index.jsx
```diff
@@ -1,26 +1,39 @@
...
 
+    beforeEach(() => {
+        global.MutationObserver = class {
+            constructor() {}
+            disconnect() {}
+            observe() {}
+        };
+    });
+
     afterEach(() => sandbox.restore());
```
KuznetsovRoman, line 19: Lets remove `global.MutationObserver` in `afterEach` in order to not pollute `global`

KuznetsovRoman (approved):
Looks like `react-testing-library` does not give a nice way to query by class name as they don't want us to do it: 

> Note that using this as an escape hatch to query by class or id is not recommended because they are invisible to the user

> If you find yourself using `container` to query for rendered elements then you should reconsider! The other queries are designed to be more resilient to changes that will be made to the component you're testing

It does not seem to worth fixing these in old tests, though.

lib/static/components/modals/screenshot-accepter/index.jsx
shadowusr: Because it wasn't quite right before. The method used state in it, and was called right after updating the state. But state is not *guaranteed* to be updated after the `setState` line. It may happen later. This became even more apparent in tests.

package.json
shadowusr: Only v3 supports native ref forwarding: https://github.com/Semantic-Org/Semantic-UI-React/issues/4438

test/func/tests/plugins/tests-menu-bar-plugin.testplane.js
shadowusr: Unfortunately, yes =(
In my tests, 500 wasn't enough. It's very annoying and weird, but I don't want to spend time on this right now

lib/static/components/header/summary/dbBtn.jsx
shadowusr: Semantic UI supports ref forwarding too, as of v3

PR #583, opened 2024-07-31
fix: do not lose meta.url if it is specified and apply base host to all urls in meta

shadowusr:
What's done:
- fixed a critical bug in suites rendering (not updating section height)
- now base host is applied to all urls in meta info
- now we respect meta.url when writing it to DB (previously if some other plugins modified meta.url, we just lost it before writing to db and used suite.url instead)


lib/static/components/section/body/meta-info/content.jsx
```diff
@@ -95,8 +95,19 @@ class MetaInfoContent extends Component {
             formattedMetaInfo[key] = {content: formattedMetaInfo[key]};
         });
 
-        if (result.suiteUrl) {
-            formattedMetaInfo.url = {content: result.metaInfo.url || result.suiteUrl, url: getUrlWithBase(result.suiteUrl, baseHost)};
+        for (const [key, value] of Object.entries(formattedMetaInfo)) {
```
DudaGod, line 98: Hm, I'm not sure that we should replace base host for all urls in meta. I believe that by default it should only be used for `url` field and other fields should be specified by user through `metaInfoBaseUrls` option.

DudaGod (approved):
🔥

PR #584, opened 2024-08-01
fix: do not fail when extend error with snippet

DudaGod:


lib/common-utils.ts
```diff
@@ -331,7 +331,7 @@ const cloneError = <T extends Error>(error: T): T => {
 };
 
 export const mergeSnippetIntoErrorStack = <T extends Error>(error: T & { snippet?: string }): T => {
-    if (!error.snippet) {
+    if (!error || !error.snippet) {
```
DudaGod, line 334: the problem was reproduced by the user a couple of times, but i can't reproduce it on the same project. So add this condition in order to not fail here anymore.

DudaGod:
It looks like there is nothing to review, so I merge it.

PR #585, opened 2024-08-02
fix: display a human-readable error message when report is completely empty

shadowusr:
What it looks like:

### DIFF
--- a/lib/static/components/suites.jsx
+++ b/lib/static/components/suites.jsx
@@ -112,7 +112,16 @@ function Suites(props) {
 
     const {visibleRootSuiteIds} = props;
 
-    if (isEmpty(visibleRootSuiteIds)) {
+    if (props.isReportEmpty) {
+        return <div className="sections sections_type_empty">
+            There are no tests to show. Please, try the following:
+            <ul>
+                <li>Check if your project contains any tests;</li>
+                <li>Check if the tool you are using is configured correctly and is able to find your tests;</li>
+                <li>Check if some critical error has occurred in logs that prevented HTML reporter from collecting any results.</li>
+            </ul>
+        </div>;
+    } else if (isEmpty(visibleRootSuiteIds)) {
         const selectedFiltersMsgs = _getSelectedFiltersInfo();
 
         return <div className="sections sections_type_empty">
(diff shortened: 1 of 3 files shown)


lib/static/components/suites.jsx
```diff
@@ -112,7 +112,16 @@ function Suites(props) {
...
     const {visibleRootSuiteIds} = props;
 
-    if (isEmpty(visibleRootSuiteIds)) {
+    if (props.isReportEmpty) {
+        return <div className="sections sections_type_empty">
+            There are no tests to show. Please, try the following:
+            <ul>
+                <li>Check if your project contains any tests;</li>
+                <li>Check if the tool you are using is configured correctly and is able to find your tests;</li>
+                <li>Check if some critical error has occurred in logs that prevented HTML reporter to collect any results.</li>
```
Kabir-Ivan, line 121: I would instead write:
"Check if some critical error has occurred in logs that prevented HTML reporter from collecting any results"

lib/static/components/suites.jsx
```diff
@@ -112,7 +112,16 @@ function Suites(props) {
 
     const {visibleRootSuiteIds} = props;
 
-    if (isEmpty(visibleRootSuiteIds)) {
+    if (props.isReportEmpty) {
+        return <div className="sections sections_type_empty">
```
KuznetsovRoman, line 116: I would pull this `html` (as well as "There are no tests that match to selected filters" html) out to functions, so here it would look like:

```js
if (props.isReportEmpty) {
    return renderEmptyReport();
}

if (isEmpty(visibleRootSuiteIds)) {
  return renderEmptyVisibleRootSuiteIds(_getSelectedFiltersInfo());
}

return (
  <div className="sections">
  ...
)

KuznetsovRoman (approved):
We have a test `should not render List component if there are no visible root suite ids` in  `test/unit/lib/static/components/suites.jsx` (line 51-57) that checks rendered text.

Let's add the same test for this case, when report is empty

lib/static/components/suites.jsx
shadowusr: IMO it only sacrifices readability without solving any actual problems. Keeping it as is for now.

PR #589, opened 2024-08-09
fix: strip ansi markup for error message

sipayRT:


test/func/tests/common/test-results-appearance.testplane.js
shadowusr, line 1: We don't need those awaits as far as I can tell... But of course they won't break nothing.

test/func/fixtures/testplane/failed-describe.testplane.js
```diff
@@ -29,6 +29,10 @@ describe('failed describe', function() {
         throw new Error('Some error');
     });
 
+    it('failed test with ansi markup', async () => {
+        await expect({a: {b: 'c'}}).toMatchObject({c: {b: 'a'}});
```
KuznetsovRoman, line 33: What kind of test is it?

KuznetsovRoman, line 33: Do we need await here?

shadowusr (approved):

KuznetsovRoman (approved):

test/func/fixtures/testplane/failed-describe.testplane.js
sipayRT: actually no, but I added it for consistency

PR #592, opened 2024-08-18
feat: implement filtering by status in new ui

shadowusr:
Example report: https://nda.ya.ru/t/VLqVLf-d77fFpu

⚠️ This PR is based on #590, only the last commit needs review (`feat: implement filtering by status in new ui`).

### DIFF
--- a/lib/static/new-ui/features/suites/components/TestStatusFilter/index.module.css
+++ b/lib/static/new-ui/features/suites/components/TestStatusFilter/index.module.css
@@ -1,0 +1,9 @@
+.test-status-filter-option {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
+    .test-status-filter-option__count {
+        margin-left: var(--g-spacing-1);
+    }
--- a/lib/static/new-ui/features/suites/components/TestStatusFilter/index.module.css
+++ b/lib/static/new-ui/features/suites/components/TestStatusFilter/index.module.css
@@ -4,6 +4,6 @@
     justify-content: center;
 }
 
-    .test-status-filter-option__count {
-        margin-left: var(--g-spacing-1);
-    }
+.test-status-filter-option__count {
+    margin-left: var(--g-spacing-1);
+}
--- a/lib/static/new-ui/features/suites/components/TestStatusFilter/index.tsx
+++ b/lib/static/new-ui/features/suites/components/TestStatusFilter/index.tsx
@@ -29,11 +29,11 @@ interface TestStatusFilterProps {
 
 function TestStatusFilterInternal({statusCounts, actions, viewMode}: TestStatusFilterProps): ReactNode {
     return <RadioButton width={'max'} onChange={(e): void => void actions.changeViewMode(e.target.value)} value={viewMode}>
-        <RadioButton.Option value={ViewMode.ALL} content={<TestStatusFilterOption status={'total'} count={statusCounts.total}/>} />
-        <RadioButton.Option value={ViewMode.PASSED} content={<TestStatusFilterOption status={TestStatus.SUCCESS} count={statusCounts.success}/>} />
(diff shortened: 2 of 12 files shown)


lib/static/new-ui/features/suites/components/TestStatusFilter/index.module.css
```diff
@@ -0,0 +1,9 @@
+.test-status-filter-option {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
+    .test-status-filter-option__count {
```
sipayRT, line 7: ←

sipayRT: or you can use stylus:

```css
.test-status-filter-option {

    &__count {
    
    }
}
```

lib/static/new-ui/features/suites/components/TestStatusFilter/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+interface TestStatusFilterProps {
+    statusCounts: StatusCounts;
+    actions: typeof actions;
+    viewMode: ViewMode;
+}
+
+function TestStatusFilterInternal({statusCounts, actions, viewMode}: TestStatusFilterProps): ReactNode {
+    return <RadioButton width={'max'} onChange={(e): void => void actions.changeViewMode(e.target.value)} value={viewMode}>
+        <RadioButton.Option value={ViewMode.ALL} content={<TestStatusFilterOption status={'total'} count={statusCounts.total}/>} />
```
sipayRT, line 32: i would like to add a title for each option (but i'm not sure that it's possible for this component)

lib/static/new-ui/features/suites/components/TestStatusFilter/selectors.ts
```diff
@@ -0,0 +1,52 @@
...
+export const getStatusCounts = createSelector(
+    [getResults, getAllBrowserIds],
+    (results, browserIds) => {
+        const latestAttempts: Record<string, {attempt: number; status: string, timestamp: number}> = {};
+        const retriedTests = new Set<string>();
+
+        let retries = 0;
+        Object.values(results).forEach(result => {
+            const {parentId: testId, attempt, status, timestamp} = result;
+            if (attempt > 0) {
```
sipayRT, line 23: can we check it by `status === RETRIED`?

lib/static/new-ui/features/suites/components/TestStatusFilter/selectors.ts
```diff
@@ -0,0 +1,52 @@
...
+            success: 0,
+            fail: 0,
+            skipped: 0,
+            total: browserIds.length,
+            retried: retriedTests.size,
+            retries,
+            idle: 0
+        };
+
+        Object.values(latestAttempts).forEach(({status}) => {
```
sipayRT, line 44: why can't we do this for all results without `retried` filtering?

sipayRT (approved):

lib/static/new-ui/features/suites/components/TestStatusFilter/selectors.ts
shadowusr: No, status stores an actual test status like fail, error or success. There's no such thing as "retried" status in sqlite db or anywhere else.

lib/static/new-ui/features/suites/components/TestStatusFilter/selectors.ts
shadowusr: We need to compute final test results before counting. Results are not linked between each other and this is the only way to determine their final status.

lib/static/new-ui/features/suites/components/TestStatusFilter/index.tsx
shadowusr: Done!
Looks like this now:

PR #595, opened 2024-08-20
fix: fix flickering when expanding/collapsing sections

shadowusr:
Test report: https://nda.ya.ru/t/OhIDE9f-77hGkc


lib/static/components/details.jsx
```diff
@@ -1,89 +1,81 @@
...
+            <Disclosure className={className} onUpdate={handleClick} size='l' expanded={isOpened}>
+                <Disclosure.Summary>
+                    {(props, defaultButton) => (
+                        <div className={classNames(className, 'details__summary')} aria-controls={props.ariaControls} onClick={props.onClick} id={props.id}>
+                            <div className="details__expand-button" onClick={e => e.stopPropagation()}>
+                                {defaultButton}
                             </div>
-                        )}
-                    </Disclosure.Summary>
-                    {type === 'image' ? this._renderContent() :
```
sipayRT, line 81: why don't you need this condition now?

lib/static/components/measurement-context.js
```diff
@@ -0,0 +1,3 @@
+import {createContext} from 'react';
+
+export const MeasurementContext = createContext({measure: () => {}});
```
sipayRT, line 3: can't figure out how it works and fixes the problem

lib/static/components/details.jsx
shadowusr: We wrap components in Card directly where we need it and pass already wrapped components as children.

The Details component shouldn't be aware of anything regarding its children IMO. It shouldn't care what it will render, it just receives `ReactNode` and renders it as-is. That makes logic a lot easier to understand and reduces complexity. That's from the software design point of view.

Another reason why I got rid of it, because it cause another visual glitch — upon collapsing Meta/History, you'd first see a frame where Disclosure contains empty card and only afterwards it collapses completely.

lib/static/components/measurement-context.js
shadowusr: **Intro**: react-virtualized needs to know exact sizes of each element for correct positioning. To let react-virtualized know the exact dimensions of element, you should use `measure()` function that is provided by `<CellMeasurer/>`. 

**Problem** (how it was before): when we expand section, it immediately renders expanded section. At this point react-virtualized knows nothing about expanded section, because `measure()` function wasn't called. On the next frame ResizeObserver detects a change and calls `measure()` that adjusts list items positioning. User sees intermediate frame and perceives it as glitch.

**Solution** (how it is now): we use `useLayoutEffect` hook inside elements that change size instead of relying solely on ResizeObserver. Now, when we expand section, the component re-renders, but before painting on the screen `useLayoutEffect` is called, where we call `measure()` to let react-virtualized adjust positioning *before* user sees anything.

Since there are many components that need to be able to call `measure()` before rendering (section, browser section, screenshot state, meta info, etc.) I introduced `MeasurementContext` just to avoid prop drilling. Architecture-wise it may be far from perfect (ideally components shouldn't be aware of this), but I think it's good enough before we switch to new UI.

Now, everything looks 100% correct and user never sees intermediate state. However there's definitely some room for optimisation in terms of performance. If you click on expand/collapse all under the devtools CPU x6 slower throttling, it takes around 5s to render. That's because measuring happens tens of times instead of just once before final render. I wasn't able to fix this quickly, and without throttling everything works fine, so decided to keep it as is.

PR #596, opened 2024-08-28
fix: pwt static image accepter

KuznetsovRoman:


lib/static/modules/reducers/tree/helpers.js
```diff
@@ -110,9 +110,9 @@ export function updateImagesStatus({tree, view, images, imageIdsArray, newStatus
             .map(imageId => getUpdatedProperty(tree, diff, ['images', 'byId', imageId, 'status']));
 
         const finalStatus = determineFinalStatus(imageStatuses);
-        const trueResultStatus = isAssertViewError(result.error) && finalStatus === ERROR ? FAIL : finalStatus;
+        const trueResultStatus = isImageError(result.error) && finalStatus === ERROR ? FAIL : finalStatus;
 
-        if (isAssertViewError(result.error) && trueResultStatus !== result.status) {
+        if (isImageError(result.error) && trueResultStatus !== result.status) {
```
KuznetsovRoman, line 115: In Testplane, we have a contract: if test result has image errors, test result has "AssertViewError", and images can have either `ImageDiffError` or `NoRefImageError`.

In Playwright adapter, though, we can have `ImageDiffError` or `NoRefImageError` as result.error. Looks like we could encounter other issues because of that.

lib/adapters/test-result/playwright.ts
```diff
@@ -34,6 +34,8 @@ export type PwtImageDiffError = ExtendedError<{snapshotName: string, diffCluster
 
 export type PwtNoRefImageError = ExtendedError<{snapshotName: string}>;
 
+export type PlaywrightImageFile = ImageFile & { relativePath: string };
```
KuznetsovRoman, line 37: I don't want to pollute "ImageFile", which is used in Testplane and generally has no guaranteed "relativePath" property

sipayRT (approved):

PR #597, opened 2024-08-30
feat: show basic test info in new ui

shadowusr:
✅ Test report available here: https://nda.ya.ru/t/kpaVBZbS77tbQ6

### What's done?
- Test title for each test, with ability to select it and copy correctly
- Attempt picker, refactored and used in both old and new UI
- Meta info, re-written from ground-up. Used in both old and new UI
- Counters in test status picker are updated according to other filters (e.g. browser or test name)
- Minor visual fixes


lib/constants/test-statuses.ts
```diff
@@ -18,7 +18,9 @@ export enum TestStatus {
     /**
      * @note used in new UI only for rendering icons
      */
-    RETRY = 'retry'
+    RETRY = 'retry',
+    /** @note used to display tests that have both failed screenshots and errors */
+    FAIL_ERROR = 'fail_error',
```
KuznetsovRoman, line 23: This is for new ui only, right?

lib/static/components/modals/screenshot-accepter/meta.jsx
```diff
@@ -1,7 +1,7 @@
 import React, {Component, Fragment} from 'react';
 import PropTypes from 'prop-types';
 
-import MetaInfoContent from '../../section/body/meta-info/content';
+import {MetaInfo as MetaInfoContent} from '@/static/new-ui/components/MetaInfo';
```
KuznetsovRoman, line 4: Are you planning to "move everything necessary to new-ui" and then "remove everything else"?

lib/static/modules/reducers/grouped-tests/helpers.js
```diff
@@ -14,7 +14,7 @@ export function handleActiveResults({tree = {}, viewMode = ViewMode.ALL, filtere
         const browser = tree.browsers.byId[result.parentId];
         const testName = browser.parentId;
 
-        if (!isTestNameMatchFilters(testName, testNameFilter, strictMatchFilter)) {
+        if (!isTestNameMatchFilters(testName, browser.name, testNameFilter, strictMatchFilter)) {
```
KuznetsovRoman, line 17: Why?

lib/static/modules/reducers/suites-page.ts
```diff
@@ -1,15 +1,22 @@
...
+        case actionNames.SUITES_PAGE_SET_SECTION_EXPANDED: {
+            const newState = Object.assign({}, state);
+            const newExpandedSectionsById = Object.assign({}, state.ui.suitesPage.expandedSectionsById, {
+                [action.payload.sectionId]: action.payload.isExpanded
+            });
+            newState.ui = Object.assign({}, state.ui, {
+                suitesPage: Object.assign({}, state.ui.suitesPage, {expandedSectionsById: newExpandedSectionsById})
+            });
+
+            return newState;
```
KuznetsovRoman, line 18: jfiy: you could do this like this:
```ts
import {set} from 'lodash';
import {applyStateUpdate} from '../utils';

const diff = {}; // difference between current state and the new one
set(
    diff,
    ["ui", "suitesPage", "expandedSectionsById", action.payload.sectionId], 
    action.payload.isExpanded
);

return applyStateUpdate(state, diff);
```

Or like this:
```ts
const diff = set(
    {},
    ["ui", "suitesPage", "expandedSectionsById", action.payload.sectionId], 
    action.payload.isExpanded
);

return applyStateUpdate(state, diff);
```

You can read more about `applyStateUpdate` and other redux helpers at https://github.com/gemini-testing/html-reporter/blob/master/lib/static/modules/utils/state.js#L3-L23

lib/static/new-ui/types/store.ts
```diff
@@ -1,51 +1,69 @@
...
     timestamp: number;
+    metaInfo: Record<string, string>;
+    suiteUrl?: string;
 }
 
 export interface ResultEntityError extends ResultEntityCommon {
-    error: Error;
-    status: TestStatus.ERROR;
+    error: TestError;
+    status: TestStatus.ERROR | TestStatus.FAIL;
```
KuznetsovRoman, line 45: Couldn't it also be "FAIL_ERROR" which you added in this PR?

lib/static/new-ui/types/store.ts
```diff
@@ -59,35 +77,54 @@ export interface SuiteState {
 
 export interface BrowserState {
     shouldBeShown: boolean;
+    retryIndex: number;
+    // True if test is not shown because of its status. Useful when computing counts by status.
+    isHiddenBecauseOfStatus: boolean;
```
KuznetsovRoman, line 82: Since you are only using "true" or not defining the property at all, maybe we should describe it as `isHiddenBecauseOfStatus?: true`?

test/func/tests/package.json
```diff
@@ -3,18 +3,18 @@
...
-    "testplane:testplane-common": "PROJECT_UNDER_TEST=testplane SERVER_PORT=8061 npx testplane --set common",
-    "testplane:testplane-eye": "PROJECT_UNDER_TEST=testplane-eye SERVER_PORT=8062 npx testplane --set eye",
-    "testplane:testplane-gui": "PROJECT_UNDER_TEST=testplane-gui SERVER_PORT=8063 npx testplane --no --set common-gui",
+    "gui:plugins": "TOOL=testplane PROJECT_UNDER_TEST=plugins SERVER_PORT=8084 npx testplane --set plugins gui",
+    "gui:testplane-tinder": "TOOL=testplane PROJECT_UNDER_TEST=testplane-gui SERVER_PORT=8084 npx testplane --set common-tinder gui",
+    "testplane:testplane-common": "TOOL=testplane PROJECT_UNDER_TEST=testplane SERVER_PORT=8061 npx testplane --set common",
+    "testplane:testplane-eye": "TOOL=testplane PROJECT_UNDER_TEST=testplane-eye SERVER_PORT=8062 npx testplane --set eye",
+    "testplane:testplane-gui": "TOOL=testplane PROJECT_UNDER_TEST=testplane-gui SERVER_PORT=8063 npx testplane --no --set common-gui",
     "testplane:playwright": "PROJECT_UNDER_TEST=playwright SERVER_PORT=8065 npx testplane --set common",
```
KuznetsovRoman, line 15: Maybe we should provide `TOOL=playwright` to pwt tests too?
If i saw `if (process.env.TOOL == "testplane")`, i wouldn't expect it to be `undefined` for pwt tests

KuznetsovRoman (approved):

lib/constants/test-statuses.ts
shadowusr: No, it's used in AttemptPicker, which is used both in new UI and old UI.

lib/static/components/modals/screenshot-accepter/meta.jsx
shadowusr: Yeah, that's the idea for components. I currently do not touch reducers and store-related logic.

lib/constants/test-statuses.ts
KuznetsovRoman: If you want to put this status here, let's also mention "it could only be used in AttemptPicker", because all the other test results are really could be seen in testResult.status

lib/static/modules/reducers/grouped-tests/helpers.js
shadowusr: To be able to search for the test including browser name. Because in new UI when you click on copy, the whole title is copied, including browser name.

lib/static/new-ui/types/store.ts
shadowusr: No it can't FAIL_ERROR is a status that's added by html-reporter, but we receive results from testplane

lib/static/new-ui/types/store.ts
shadowusr: Made this field optional. But I see no reason for using `true` instead of `boolean`. Right now it's either true or absent, but why prohibit setting it to false.

lib/static/modules/reducers/suites-page.ts
shadowusr: Yeah, thanks, that's much more convenient. Ideally this should be handled by Immer of course =)

PR #598, opened 2024-09-06
fix(gui): wrong base url for tests in gui mode

sipayRT:


lib/common-utils.ts
```diff
@@ -72,21 +72,23 @@ export const determineFinalStatus = (statuses: TestStatus[]): TestStatus | null
 
 export const getUrlWithBase = (url: string | undefined, base: string | undefined): string => {
     try {
+        if (isEmpty(base)) {
```
DudaGod, line 75: It looks like that we can write this code outside `try` statement, because i don't see what can trigger an error in this code.

lib/static/components/icons/view-in-browser/index.jsx
```diff
@@ -41,7 +41,7 @@ class ViewInBrowser extends Component {
         return (
             <a
                 className={className}
-                href={getUrlWithBase(getRelativeUrl(suiteUrl), baseHost)}
+                href={getUrlWithBase(suiteUrl, baseHost)}
```
DudaGod, line 44: what was the problem with using `getRelativeUrl` method?

DudaGod (approved):

lib/static/components/icons/view-in-browser/index.jsx
sipayRT: in case of empty baseHost relative path will be returned:

```
getUrlWithBase(getRelativeUrl("https://ya.ru/some/path"), null); ==> /some/path
```

and such relative paths will be resolved in the browser relative to the report host

PR #600, opened 2024-09-14
refactor: re-write screenshots viewing to using css only

shadowusr:
Demo report (large project): https://nda.ya.ru/t/fUzVGiOZ78FcQg
Demo report (images of different sizes and scales): https://nda.ya.ru/t/f-nAg8Oy78FcXq

Note: no changes in new UI are present. This is a prerequisite step to adding screenshots viewer in new UI.

PR #601, opened 2024-09-16
chore(new-ui): implement screenshots display

shadowusr:
⚠️ This PR is based on #600. **Only last commit needs review.**

Test reports:
- With various screenshot states: https://nda.ya.ru/t/G-eslI5L78GALr
- Large project: https://nda.ya.ru/t/gGn-IsIl78GAgR

### DIFF
--- a/lib/constants/diff-modes.ts
+++ b/lib/constants/diff-modes.ts
@@ -3,38 +3,38 @@ import {ValueOf} from 'type-fest';
 export const DiffModes = {
     THREE_UP: {
         id: '3-up',
-        title: '3-up',
-        description: 'images in column'
+        title: 'List',
+        description: 'List. Show images one after another in vertical layout.'
     },
     THREE_UP_SCALED: {
         id: '3-up-scaled',
-        title: '3-up scaled',
-        description: 'scaled images in row'
+        title: 'SbS',
+        description: 'Side by Side. Show images in one row.'
     },
     THREE_UP_SCALED_TO_FIT: {
         id: '3-up-scaled-to-fit',
-        title: '3-up scaled to fit',
-        description: 'scaled to browser height images in row'
+        title: 'SbS (fit screen)',
+        description: 'Side by Side. Show images in one row and scale them down if needed to fit the screen.'
     },
     ONLY_DIFF: {
         id: 'only-diff',
-        title: 'Only diff',
-        description: 'click on image to see area with diff'
+        title: 'Only Diff',
+        description: 'Only Diff. Show only diff image, click to highlight diff areas.'
     },
     SWITCH: {
         id: 'switch',
         title: 'Switch',
-        description: 'click on image to switch'
+        description: 'Switch. Click to switch between expected and actual images.'
     },
     SWIPE: {
         id: 'swipe',
         title: 'Swipe',
-        description: 'move divider'
+        description: 'Swipe. Move the divider to compare expected and actual images.'
     },
     ONION_SKIN: {
         id: 'onion-skin',
         title: 'Onion skin',
-        description: 'move slider'
+        description: 'Onion Skin. Change the image opacity to compare expected and actual images.'
     }
 } as const;
 
--- a/lib/static/new-ui/components/AssertViewResult/index.module.css
+++ b/lib/static/new-ui/components/AssertViewResult/index.module.css
@@ -1,0 +1,17 @@
+.diff-viewer-container {
+    display: flex;
+    flex-direction: column;
+    padding-left: calc(var(--indent) * 24px);
+    padding-right: 1px
+}
+
+.diff-mode-switcher {
+    --g-color-base-background: #fff;
+    margin: 12px auto;
+}
+
+.screenshot {
+    margin: 8px 0;
+    padding-left: calc(var(--indent) * 24px);
+    padding-right: 1px;
+}
--- a/lib/static/new-ui/components/DiffViewer/index.module.css
+++ b/lib/static/new-ui/components/DiffViewer/index.module.css
@@ -1,0 +1,8 @@
+.image-label, .image-label + div {
+    margin-bottom: 8px;
+}
+
+.image-label-subtitle {
+    color: var(--g-color-private-black-400);
+    margin-left: 4px;
+}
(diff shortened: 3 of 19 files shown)


lib/constants/diff-modes.ts
```diff
@@ -3,38 +3,38 @@ import {ValueOf} from 'type-fest';
...
-        title: 'Only diff',
-        description: 'click on image to see area with diff'
+        title: 'Only Diff',
+        description: 'Only Diff. Show only diff image, click to highlight diff areas.'
     },
     SWITCH: {
         id: 'switch',
         title: 'Switch',
-        description: 'click on image to switch'
+        description: 'Switch. Click to switch between expected and actual images.'
```
sipayRT, line 27: I think we don't need to duplicate the name here

lib/static/new-ui/components/AssertViewResult/index.module.css
```diff
@@ -0,0 +1,17 @@
...
+
+.diff-mode-switcher {
+    --g-color-base-background: #fff;
+    margin: 12px auto;
+}
+
+.screenshot {
+    margin: 8px 0;
+    padding-left: calc(var(--indent) * 24px);
+    padding-right: 1px;
```
sipayRT, line 16: can't understand why we need this indentation here and in the `diff-viewer-container` block

lib/static/new-ui/components/DiffViewer/index.module.css
```diff
@@ -0,0 +1,8 @@
+.image-label, .image-label + div {
```
sipayRT, line 1: we need some classname for this `div`

lib/static/new-ui/components/DiffViewer/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+            return <SwitchMode expected={expectedImg} actual={actualImg} />;
+
+        case DiffModes.SWIPE.id:
+            return <SwipeMode expected={expectedImg} actual={actualImg} />;
+
+        case DiffModes.ONION_SKIN.id:
+            return <OnionSkinMode expected={expectedImg} actual={actualImg} />;
+
+        case DiffModes.THREE_UP_SCALED.id:
+            return <SideBySideMode actual={actualImg} diff={diffImg} expected={expectedImg} />;
```
sipayRT, line 71: `expected, actual, diff` please

lib/static/new-ui/components/DiffViewer/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+        case DiffModes.THREE_UP_SCALED_TO_FIT.id: {
+            const desiredHeight = props.desiredHeight ?? 'calc(100vh - 180px)';
+
+            return <SideBySideToFitMode desiredHeight={desiredHeight} actual={actualImg} diff={diffImg} expected={expectedImg} />;
```
sipayRT, line 76: and here

lib/static/new-ui/components/DiffViewer/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+        }
+        case DiffModes.THREE_UP.id:
+        default:
+            return <ListMode actual={actualImg} diff={diffImg} expected={expectedImg} />;
```
sipayRT, line 80: for the last time, I swear :)

lib/static/new-ui/features/suites/components/SuitesPage/index.module.css
```diff
@@ -46,6 +46,6 @@
     background-color: white;
     position: sticky;
     top: 0;
-    z-index: 1;
+    z-index: 10;
```
sipayRT, line 49: hm?

lib/static/new-ui/features/suites/components/TestSteps/index.tsx
```diff
@@ -55,15 +58,15 @@ function TestStepsInternal(props: TestStepsProps): ReactNode {
                 const item = items.structure.itemsById[itemId];
 
                 if (item.type === StepType.Action) {
-                    const shouldHighlightFail = item.status === TestStatus.ERROR && !item.isGroup;
+                    const shouldHighlightFail = (item.status === TestStatus.ERROR || item.status === TestStatus.FAIL) && !item.isGroup;
```
sipayRT, line 61: its time to merge this check to the `isFailed` function

lib/static/styles.css
```diff
@@ -979,6 +979,6 @@ a:active {
     color: var(--g-color-private-cool-grey-600-solid);
 }
 
-.icon-retry {
+.icon-retry, .icon-updated {
```
sipayRT, line 982: why are these colors the same?

sipayRT (approved):

lib/constants/diff-modes.ts
shadowusr: I think we do, because for instance we have SbS label and Side by Side as an explanation. Keeping it for now.

lib/static/new-ui/features/suites/components/SuitesPage/index.module.css
shadowusr: DiffViewer has absolutely positioned elements, so z-index 1 is no longer enough for the sticky-header.

lib/static/styles.css
shadowusr: Now it's pink:

Screenshot 2024-09-17 at 19 00 14

lib/static/new-ui/components/DiffViewer/index.module.css
shadowusr: Problem is that we have the following structure:

```
<DiffViewer>
  <Label/>
  <SideBySIdeMode>
    <Screenshot>
      <div></div>      <—— div that we want to target
    </Screenshot>
  </SideBySIdeMode>
</DiffViewer>
```

So the target div from the react point of view is a few levels deeper than the label and we can't easily set class name in DiffViewer for that. Setting global class name inside screenshot sounds wrong too. So IMO `+ div` is the best choice out of not-so-great options.

lib/static/new-ui/components/AssertViewResult/index.module.css
shadowusr: Because screenshots are nested in steps tree, namely inside assertView command which can be inside some other command, e.g. runStep. These paddings are to maintain correct indents.

PR #605, opened 2024-10-02
feat(new-ui): add ability to accept and undo screenshots

shadowusr:
⚠️ Only **last three commits** need review.

What's done:
- Ability to accept/undo screenshots
- Tweaks to tinder mode, including compact attempt picker and status display
- Other minor fixes

Test report: https://nda.ya.ru/t/TDDrfcZG78eWpd

### DIFF
--- a/lib/constants/features.ts
+++ b/lib/constants/features.ts
@@ -5,3 +5,7 @@ export interface Feature {
 export const RunTestsFeature = {
     name: 'run-tests'
 } as const satisfies Feature;
+
+export const EditScreensFeature = {
+    name: 'edit-screens'
+} as const satisfies Feature;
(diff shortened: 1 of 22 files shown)


DudaGod (commented):
Current questions:
- diff inside Visual checks looks like has incorrect dimension (example: `Diff1192034px ⋅ 99.5%`)
- should we render in Visual checks retries which doesn't fail with image comparison?

DudaGod (approved):

shadowusr:
> diff inside Visual checks looks like has incorrect dimension

It's actually correct, it contains not image dimensions but amount of pixels and % that are different.

> should we render in Visual checks retries which doesn't fail with image comparison?

This question we talked about privately as well and agreed that for now we should keep current behavior:
- Speaking of retries that don't have such image at all, this case is very rare and displaying all retries adds consistency to the UI
- Considering showing successful retries in general, we keep it for the same reason of consistency & in the future we'll have filtering capabilities on the left

PR #606, opened 2024-10-03
fix: correctly build link to error details

DudaGod:
### What is done

It turned out that after refactoring, `extractErrorDetails` is called several times - https://github.com/gemini-testing/html-reporter/blob/v10.7.0/lib/adapters/test-result/utils/index.ts#L50-L64 which builds the path to the file and calls `Date.now()` inside it. As a result, each call generates a unique path. And one is used for link inside html-reporter and another for save file in folder.

Now we will use the already existing value.

### DIFF
--- a/lib/adapters/test-result/transformers/tree.ts
+++ b/lib/adapters/test-result/transformers/tree.ts
@@ -2,7 +2,6 @@ import {ReporterTestResult} from '../index';
 import _ from 'lodash';
 import {BaseTreeTestResult} from '../../../tests-tree-builder/base';
 import {DbTestResultTransformer} from './db';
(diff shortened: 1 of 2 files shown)


lib/adapters/test-result/transformers/tree.ts
```diff
@@ -21,7 +20,7 @@ export class TreeTestResultTransformer {
         return {
             ..._.omit(result, 'imagesInfo'),
             attempt: testResult.attempt,
-            errorDetails: extractErrorDetails(testResult)
+            errorDetails: testResult.errorDetails
```
DudaGod, line 23: We don't have tests for these files

sipayRT (approved):

PR #609, opened 2024-10-16
feat: implement a button to switch between new and old UI

shadowusr:
### What's done:
- Go to new UI button in old UI
- Settings panel with go back to old UI
- The choice of UI mode is remembered

Demo report: https://nda.ya.ru/t/dW1Fx8qe78xojb
Demo GIF: https://nda.ya.ru/t/Il_hVvfw78xoku

### DIFF
--- a/lib/static/new-ui/app/App.tsx
+++ b/lib/static/new-ui/app/App.tsx
@@ -2,10 +2,9 @@ import {ThemeProvider} from '@gravity-ui/uikit';
 import React, {ReactNode, StrictMode} from 'react';
 import {MainLayout} from '../components/MainLayout';
 import {HashRouter, Navigate, Route, Routes} from 'react-router-dom';
-import {CircleInfo, Eye, ListCheck} from '@gravity-ui/icons';
+import {Eye, ListCheck} from '@gravity-ui/icons';
 import {SuitesPage} from '../features/suites/components/SuitesPage';
 import {VisualChecksPage} from '../features/visual-checks/components/VisualChecksPage';
-import {InfoPage} from '../features/info/components/InfoPage';
 
 import '@gravity-ui/uikit/styles/fonts.css';
 import '@gravity-ui/uikit/styles/styles.css';
@@ -23,8 +22,7 @@ export function App(): ReactNode {
             element: <SuitesPage/>,
             children: [<Route key={'suite'} path=':suiteId' element= {<SuitesPage/>} />]
         },
-        {title: 'Visual Checks', url: '/visual-checks', icon: Eye, element: <VisualChecksPage/>},
-        {title: 'Info', url: '/info', icon: CircleInfo, element: <InfoPage/>}
+        {title: 'Visual Checks', url: '/visual-checks', icon: Eye, element: <VisualChecksPage/>}
     ];
 
     return <StrictMode>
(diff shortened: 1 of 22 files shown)


lib/static/new-ui/app/App.tsx
```diff
@@ -23,8 +22,7 @@ export function App(): ReactNode {
             element: <SuitesPage/>,
             children: [<Route key={'suite'} path=':suiteId' element= {<SuitesPage/>} />]
         },
-        {title: 'Visual Checks', url: '/visual-checks', icon: Eye, element: <VisualChecksPage/>},
-        {title: 'Info', url: '/info', icon: CircleInfo, element: <InfoPage/>}
```
KuznetsovRoman, line 27: Why did you remove it?

KuznetsovRoman (approved):

lib/static/new-ui/app/App.tsx
shadowusr: I plan to implement info as a panel instead of a page. But right now it's empty so there's no point in misleading users, i will add it back when it'll be ready.

lib/static/new-ui/components/MainLayout/Footer.tsx
```diff
@@ -0,0 +1,67 @@
...
+import {PanelId} from '@/static/new-ui/components/MainLayout/index';
+
+interface FooterProps {
+    visiblePanel: PanelId | null;
+    onFooterItemClick: (item: GravityMenuItem) => void;
+}
+
+export function Footer(props: FooterProps): ReactNode {
+    const isInitialized = useSelector(getIsInitialized);
+    const [isHintVisible, setIsHintVisible] = useState<undefined | boolean>(undefined);
```
KuznetsovRoman, line 21: I dont like explicit undefined here
Lets use "false" or "null" instead of undefined here

lib/static/new-ui/components/UiModeHintNotification/index.module.css
```diff
@@ -0,0 +1,91 @@
...
+
+    80% {
+        filter: blur(0px);
+        transform: scale(1.02);
+    }
+
+    100% {
+        filter: blur(0px);
+        transform: scale(1);
+        visibility: visible;
```
KuznetsovRoman, line 20: Seems like "visibility: visible" does nothing here

lib/static/new-ui/components/UiModeHintNotification/index.module.css
```diff
@@ -0,0 +1,91 @@
...
+    border-radius: 10px;
+    color: rgba(255, 255, 255, .9);
+    fill: rgba(255, 255, 255, .9);
+    width: 700px;
+    display: flex;
+    gap: 8px;
+    box-shadow: 0 0 16px 0 #00000036;
+    align-items: center;
+
+    visibility: hidden;
```
KuznetsovRoman, line 58: Can we remove "visibility: hidden" from here and from "notification-appear" animation?
Feels weird, when "container" class is hidden by default (even when if it is notification hint container)

KuznetsovRoman (approved):

lib/static/new-ui/components/UiModeHintNotification/index.module.css
shadowusr: No, it's crucial here. When using animation-fill-mode forwards, setting visibility at the final stage forces element to stay visible after animation end. If we remove this line, notification will abruptly disappear as soon as animation ends.

lib/static/new-ui/components/UiModeHintNotification/index.module.css
shadowusr: We can't, discussed this in chat

PR #611, opened 2024-10-28
fix: handle invalid ref image errors

shadowusr:
invalid-ref-accept

### DIFF
--- a/lib/errors/index.ts
+++ b/lib/errors/index.ts
@@ -6,6 +6,7 @@ export const ErrorName = {
     GENERAL_ERROR: 'Error',
     IMAGE_DIFF: 'ImageDiffError',
     NO_REF_IMAGE: 'NoRefImageError',
+    INVALID_REF_IMAGE: 'InvalidRefImageError',
     ASSERT_VIEW: 'AssertViewError'
 } as const;
 export type ErrorName = ValueOf<typeof ErrorName>;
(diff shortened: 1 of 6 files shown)


lib/errors/index.ts
```diff
@@ -34,3 +35,12 @@ export interface NoRefImageError {
...
     refImg: ImageFile;
 }
+
+export interface InvalidRefImageError {
+    name: ErrorNames['INVALID_REF_IMAGE'];
+    stateName: string;
+    message: string;
+    stack?: string;
+    currImg: ImageFile;
+    refImg: ImageFile;
```
KuznetsovRoman, line 45: Why is it required, just like in "NoRefImageError"?
Feels like this property shouldn't exist.
Not important, though

lib/errors/index.ts
```diff
@@ -34,3 +35,12 @@ export interface NoRefImageError {
     currImg: ImageFile;
     refImg: ImageFile;
 }
+
+export interface InvalidRefImageError {
```
KuznetsovRoman, line 39: Lets leave a comment here: "could be thrown if testplane >= X.Y.Z", like it is described higher, so we knew its only supported in testplane X.Y.Z+

lib/errors/index.ts
shadowusr: Because errors thrown by Testplane always have this field. In this case refImg stores path to ref image that turned out to be invalid. IMO a useful piece of info.

PR #613, opened 2024-11-07
feat: implement toolbar for selecting tests and quick actions

shadowusr:
### Demo report
[Link](https://nda.ya.ru/t/xBo82_gv79VbKi)

#### New toolbar at the top of tree view
toolbar-demo

#### Handling of empty states in tree view and report
127 0 0 1_8080_new-ui (1)

localhost_8000_new-ui


lib/static/new-ui/components/Screenshot/index.tsx
```diff
@@ -21,12 +21,25 @@ interface ScreenshotProps {
...
     };
     style?: React.CSSProperties;
+    onMouseEnter?: (e: React.MouseEvent<HTMLImageElement>) => void;
+    onMouseLeave?: (e: React.MouseEvent<HTMLImageElement>) => void;
+    onMouseMove?: (e: React.MouseEvent<HTMLImageElement>) => void;
 }
 
-export function Screenshot(props: ScreenshotProps): ReactNode {
-    const imageRef = useRef<HTMLImageElement | null>(null);
+export const Screenshot = forwardRef<HTMLImageElement, ScreenshotProps>(function Screenshot(props, imageRef): ReactNode {
```
shadowusr, line 29: The essence of changes in this file: allow passing ref to img element and allow to subscribe to events.

lib/static/new-ui/components/AssertViewStatus/index.tsx
```diff
@@ -1,45 +1,12 @@
...
-    } else if (isInvalidRefImageError((image as ImageEntityError).error)) {
-        status = <><Icon data={FileExclamation} width={16}/><span>Reference is broken</span></>;
-    } else if (image.status === TestStatus.FAIL) {
-        status = <><Icon data={ArrowRightArrowLeft} width={16}/><span>Difference detected</span></>;
-    } else if (image.status === TestStatus.UPDATED) {
-        status = <><Icon data={FileCheck} width={16}/><span>Reference updated</span></>;
-    }
-
-    return <div className={styles.container}>{status}</div>;
+    return <div className={styles.container}>{getAssertViewStatusIcon(image)}<span>{getAssertViewStatusMessage(image)}</span></div>;
```
shadowusr, line 11: Moved hints & icons to helpers to be able to use them separately, e.g. in tree view.

lib/static/new-ui/features/suites/components/SuitesTreeView/index.tsx
```diff
@@ -59,39 +60,56 @@ function SuitesTreeViewInternal(props: SuitesTreeViewProps): ReactNode {
...
 
-        props.actions.setStrictMatchFilter(false);
+        dispatch(setStrictMatchFilter(false));
 
+        let timeoutId: NodeJS.Timeout;
         if (suiteId) {
-            props.actions.suitesPageSetCurrentSuite(suiteId);
-            virtualizer.scrollToIndex(list.structure.visibleFlattenIds.indexOf(suiteId), {align: 'start'});
+            dispatch(suitesPageSetCurrentSuite(suiteId));
+            timeoutId = setTimeout(() => {
```
shadowusr, line 87: This tiny timeout is needed to postpone scrolling to current item. Otherwise, when loading thousands of items, sometimes scroll position can be incorrect.

lib/static/new-ui/features/suites/components/SuitesTreeView/index.tsx
```diff
@@ -7,16 +7,12 @@ import {
 } from '@gravity-ui/uikit/unstable';
```
shadowusr, line 7: The essence of changes in this file — expose imperative handle to allow scrolling to certain item in tree and migrate to hooks redux API instead of `connect()`.

lib/static/new-ui/features/suites/components/TreeActionsToolbar/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+import {EditScreensFeature, RunTestsFeature} from '@/constants';
+
+interface TreeActionsToolbarProps {
+    onHighlightCurrentTest?: () => void;
+}
+
+export function TreeActionsToolbar(props: TreeActionsToolbarProps): ReactNode {
+    const dispatch = useDispatch();
+
+    const rootSuiteIds = useSelector((state: State) => state.tree.suites.allRootIds);
```
KuznetsovRoman, line 47: nit: it would be nice to make our own `useSelector` like this:

```ts
import {useSelector as useSelectorBase} from 'react-redux';

export const useSelector<T> = useSelectorBase as (cb: ((state: State) => T) => T);
```
So here you wouldn't need to specify state type:
```ts
const rootSuiteIds = useSelector(state => state.tree.suites.allRootIds);
```

lib/static/new-ui/features/suites/components/TreeActionsToolbar/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+        .find(feature => feature.name === EditScreensFeature.name);
+
+    const isSelectedAll = useMemo(() => {
+        for (const suiteId of rootSuiteIds) {
+            if (suitesStateById[suiteId].checkStatus !== CHECKED) {
+                return false;
+            }
+        }
+
+        return true;
```
KuznetsovRoman, line 69: Why its not `Array.every`?
```ts
return rootSuiteIds.every(suiteId => suitesStateById[suiteId].checkStatus === CHECKED);
```

lib/static/new-ui/features/suites/components/TreeActionsToolbar/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+    }, [suitesStateById, rootSuiteIds]);
+
+    const isSelectedAtLeastOne = useMemo(() => {
+        for (const suiteId of rootSuiteIds) {
+            if (suitesStateById[suiteId].shouldBeShown && (suitesStateById[suiteId].checkStatus === CHECKED || suitesStateById[suiteId].checkStatus === INDETERMINATE)) {
+                return true;
+            }
+        }
+
+        return false;
```
KuznetsovRoman, line 79: Why its not `Array.some`?
```ts
return rootSuiteIds.some(suiteId => {
    const isShown = suitesStateById[suiteId].shouldBeShown;
    const isChecked = [CHECKED, INDETERMINATE].includes(suitesStateById[suiteId].checkStatus);

    return isShown && isChecked;
});
```

lib/static/new-ui/features/suites/components/TreeActionsToolbar/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+    const selectedTestsCount = useMemo(() => {
+        let count = 0;
+
+        for (const browser of Object.values(browsersStateById)) {
+            if (browser.checkStatus === CHECKED) {
+                count++;
+            }
+        }
+
+        return count;
```
KuznetsovRoman, line 101: Why its not Array.reduce?
```ts
const browserStates = Object.values(browsersStateById);

return browserStates.reduce((acc, state) => acc + Number(state.checkStatus === CHECKED), 0);
```

lib/static/new-ui/store/selectors.ts
```diff
@@ -18,3 +18,5 @@ export const getResults = (state: State): Record<string, ResultEntity> => state.
 export const getImages = (state: State): Record<string, ImageEntity> => state.tree.images.byId;
 
 export const getIsInitialized = (state: State): boolean => state.app.isInitialized;
+export const getIsStaticImageAccepterEnabled = (state: State): boolean => state.config.staticImageAccepter.enabled;
```
KuznetsovRoman, line 21: Dont check `is static image accepter enabled` from config.
Check it from `state.staticImageAccepter.enabled` instead.

staticImageAccepter is being initialized here: https://github.com/gemini-testing/html-reporter/blob/master/lib/static/modules/reducers/static-image-accepter.js#L10-L14

And it does only when [staticImageAccepter.checkIsEnabled](https://github.com/gemini-testing/html-reporter/blob/f70e78b569773ac61cc814d342fcb25b6c6afa54/lib/static/modules/static-image-accepter.ts#L36) returns true.

The difference between `state.staticImageAccepter.enabled` and `state.config.staticImageAccepter.enabled` occurs in these situations, where `state.staticImageAccepter.enabled` would be `false`:
- gui mode. `state.staticImageAccepter.enabled` is false in GUI;
- invalid config. If config is invalid, static image accepter is turned off.

At the same time, `state.config.staticImageAccepter.enabled` has the value which user set in html-reporter config to be

lib/static/new-ui/features/suites/components/TreeActionsToolbar/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+            dispatch(retrySuite(visibleTests));
+        }
+    };
+
+    const handleUndo = (): void => {
+        const acceptableImageIds = activeImages
+            .filter(image => isScreenRevertable({image, gui: isGuiMode, isLastResult: true, isStaticImageAccepterEnabled}))
+            .map(image => image.id);
+
+        if (isStaticImageAccepterEnabled && !isGuiMode) {
```
KuznetsovRoman, line 137: If you were using `isStaticImageAccepterEnabled` as selector from `state.staticImageAccepter.enabled` instead of it being selector from `state.config`, you wouldn't need to be checking "if using gui" here (because `state.staticImageAccepter.enabled` is false in GUI)

lib/static/new-ui/store/selectors.ts
KuznetsovRoman: In fact, that was the reason i created `state.staticImageAccepter.enabled` in the first place, so there would be one single point where we could check "is there working static image accepter" and don't worry about cases, where "yes, it is enabled in config, but currently is disabled because of gui"

lib/static/new-ui/features/suites/components/TreeViewItemSubtitle/index.tsx
```diff
@@ -12,11 +16,23 @@ interface TreeViewItemSubtitleProps {
...
+                return <div key={index}>
+                    <span className={styles.imageStatus}>{(imageEntity as ImageEntityFail).stateName} ⋅ {getAssertViewStatusMessage(imageEntity)}</span>
+                    <ImageWithMagnifier image={image} style={{maxWidth: '99%', marginTop: '4px', maxHeight: '40vh'}} scrollContainerRef={props.scrollContainerRef}/>
+                </div>;
+            })}
+        </div>;
     } else if (props.item.type === TreeViewItemType.Browser && props.item.errorStack) {
         return <div className={classNames(styles['tree-view-item-subtitle__error-stack'], props.className)}>
-            {props.item.errorStack}
+            {stripAnsi(props.item.errorStack)}
```
KuznetsovRoman, line 35: What is item.errorStack?
What is this component?

lib/static/modules/reducers/static-image-accepter.js
```diff
@@ -58,14 +58,14 @@ export default (state, action) => {
...
-
             const acceptableImagesDiff = {};
             const diff = set({}, ['staticImageAccepter', 'acceptableImages'], acceptableImagesDiff);
             const imagesToCommitCount = get(state, ['staticImageAccepter', 'imagesToCommitCount']);
 
-            set(acceptableImagesDiff, [imageId, 'commitStatus'], null);
-            set(diff, ['staticImageAccepter', 'imagesToCommitCount'], imagesToCommitCount - 1);
+            for (const imageId of action.payload) {
+                set(acceptableImagesDiff, [imageId, 'commitStatus'], null);
+                set(diff, ['staticImageAccepter', 'imagesToCommitCount'], imagesToCommitCount - 1);
```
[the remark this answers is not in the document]
KuznetsovRoman: Please add corresponding test where this action is called with multiple images to unstage

KuznetsovRoman (changes requested):

lib/static/new-ui/features/suites/components/TreeActionsToolbar/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+            dispatch(staticAccepterStageScreenshot(acceptableImageIds));
+        } else {
+            dispatch(acceptOpened(acceptableImageIds));
+        }
+    };
+
+    const viewButtons = <>
+        {isRunTestsAvailable && <IconButton icon={<Icon data={Play} height={14}/>}
+            tooltip={isSelectedAtLeastOne ? 'Run selected' : 'Run visible'} view={'flat'} onClick={handleRun}
+            disabled={isRunning || !isInitialized}></IconButton>}
```
sipayRT, line 159: you can save this result in the `isDisabled` variable to reuse it below

lib/static/new-ui/utils/assert-view-status.tsx
```diff
@@ -0,0 +1,59 @@
...
+    FilePlus,
+    SquareExclamation,
+    SquareXmark
+} from '@gravity-ui/icons';
+import {TestStatus} from '@/constants';
+import {isInvalidRefImageError, isNoRefImageError} from '@/common-utils';
+import React, {ReactNode} from 'react';
+
+export const getAssertViewStatusIcon = (image: ImageEntity | null): ReactNode => {
+    if (image === null) {
```
sipayRT, line 18: switch/case?

sipayRT (approved):

lib/static/new-ui/utils/assert-view-status.tsx
KuznetsovRoman: Wouldn't work

lib/static/modules/reducers/static-image-accepter.js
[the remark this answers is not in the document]
shadowusr: Fixed and added a test. Forgot about this line because I've been changing the code a lot.

lib/static/new-ui/features/suites/components/TreeActionsToolbar/index.tsx
shadowusr: Well in fact you can augment default root state without any additional hooks. Implemented it now, so by default useSelector is aware of our types.

lib/static/new-ui/features/suites/components/TreeActionsToolbar/index.tsx
shadowusr: Pretty sure it's a matter of taste, but okay, switched it to `some()`.

lib/static/new-ui/features/suites/components/TreeActionsToolbar/index.tsx
shadowusr: Because it's easier to read in this specific case IMO. But if that really matters, changed to reduce.

lib/static/new-ui/store/selectors.ts
shadowusr: IMO having two `enabled` fields in state is pretty misleading, I'd never figured the difference if you didn't tell. The `staticImageAccepter.enabled` could be implemented as a selector named something like "shouldGuiniBeRendered" (because it can be derived from the rest of the state).

For now, changed to the field you recommended to use

lib/static/new-ui/features/suites/components/TreeActionsToolbar/index.tsx
shadowusr: Removed those.

lib/static/new-ui/features/suites/components/TreeViewItemSubtitle/index.tsx
shadowusr: TreeViewItem — component that's used for displaying nodes of the tree view. In Gravity UI, it can accept subtitle. We use subtitle to render previews of frequently needed info, e.g. error of last retry or diff of last retry.

`errorStack` is used to store error stack/snipped of last retry. I'll probably add comments to the item interface in the next PR.

lib/static/new-ui/features/suites/components/SuitesTreeView/index.tsx
```diff
@@ -59,39 +60,56 @@ function SuitesTreeViewInternal(props: SuitesTreeViewProps): ReactNode {
     const virtualizer = useVirtualizer({
         count: list.structure.visibleFlattenIds.length,
         getScrollElement: () => parentRef.current,
-        estimateSize: () => 28,
+        estimateSize: () => 32,
         getItemKey: useCallback((index: number) => list.structure.visibleFlattenIds[index], [list]),
-        overscan: 200
+        overscan: 50
```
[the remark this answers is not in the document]
shadowusr: Just by trial and error. Figured it will bring smoother experience when scrolling and working with the tree.

lib/static/new-ui/utils/assert-view-status.tsx
shadowusr: re-wrote to switch/case.

lib/static/new-ui/store/selectors.ts
KuznetsovRoman: > misleading

Yes, but "config" is just plugin config and I think it should not be modified. It's just what user set it to be.

I don't think there is anything we can do about it

PR #614, opened 2024-11-14
fix: fix diff bubbles in side-by-side fit screen mode

shadowusr:
Original user's report, with fix from this PR: https://nda.ya.ru/t/mFurBb-x79eJMr

My custom report with tons of different image aspect ratios: https://nda.ya.ru/t/zPwpfy4F79eJUT


DudaGod (approved):
🔥

PR #618, opened 2024-11-24
fix: display skip reason in new ui

shadowusr:
### What's done:
- Skip reason is now displayed as part of preview in tree view in new ui
- Skip reason is displayed in meta when possible

Demo report: https://nda.ya.ru/t/Ba_4zPC979ttwt

### DIFF
--- a/lib/static/new-ui/components/MetaInfo/index.tsx
+++ b/lib/static/new-ui/components/MetaInfo/index.tsx
@@ -1,16 +1,17 @@
 import path from 'path';
 
 import {DefinitionList} from '@gravity-ui/components';
-import {mapValues, isObject, omitBy, isEmpty} from 'lodash';
+import {isEmpty, isObject, mapValues, omitBy} from 'lodash';
 import React, {ReactNode} from 'react';
 import {connect} from 'react-redux';
 
-import {isUrl, getUrlWithBase, getRelativeUrl} from '@/common-utils';
+import {getRelativeUrl, getUrlWithBase, isUrl} from '@/common-utils';
 import {ResultEntity, State} from '@/static/new-ui/types/store';
 import {HtmlReporterValues} from '@/plugin-api';
 import {ReporterConfig} from '@/types';
 import styles from './index.module.css';
-import Parser from 'html-react-parser';
+import {makeLinksClickable} from '@/static/new-ui/utils';
+import {TestStatus} from '@/constants';
 
 const serializeMetaValues = (metaInfo: Record<string, unknown>): Record<string, string> =>
     mapValues(metaInfo, (v): string => {
@@ -21,9 +22,9 @@ const serializeMetaValues = (metaInfo: Record<string, unknown>): Record<string, 
         return v?.toString() ?? 'No value';
     });
 
-interface MetaInfoItem {
+interface MetaInfoItem<T> {
     label: string;
-    content: string;
+    content: T;
     url?: string;
     copyText?: string;
 }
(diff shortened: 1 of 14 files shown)


lib/static/new-ui/components/MetaInfo/index.tsx
```diff
@@ -109,6 +110,23 @@ function MetaInfoInternal(props: MetaInfoInternalProps): ReactNode {
         });
     }
 
+    const shouldAddSkipReason = Boolean(!metaInfoItemsWithResolvedUrls.find(item => item.label === 'muteReason'));
```
DudaGod, line 113: I would use a separate variable for this expression - `Boolean(!metaInfoItemsWithResolvedUrls.find(item => item.label === 'muteReason'))`. For example - `hasMuteReason`

lib/static/new-ui/components/MetaInfo/index.tsx
```diff
@@ -109,6 +110,23 @@ function MetaInfoInternal(props: MetaInfoInternalProps): ReactNode {
...
+                label: 'skipReason',
+                content: reason
+            });
+        } else if (!Array.isArray(reason) && typeof reason.props.children === 'string' && typeof reason.props.href === 'string') {
+            metaInfoItemsWithResolvedUrls.push({
+                label: 'skipReason',
+                content: reason.props.children,
+                url: reason.props.href
+            });
+        }
```
DudaGod, line 127: can `reason` be something else? I mean is there a condition under which the value does not fall into any of the described conditions (`if` and `else if`).

lib/static/new-ui/components/MetaInfo/index.tsx
```diff
@@ -109,6 +110,23 @@ function MetaInfoInternal(props: MetaInfoInternalProps): ReactNode {
...
 
+    const shouldAddSkipReason = Boolean(!metaInfoItemsWithResolvedUrls.find(item => item.label === 'muteReason'));
+    if (result.skipReason && shouldAddSkipReason) {
+        const reason = Parser(result.skipReason);
+        if (typeof reason === 'string') {
+            metaInfoItemsWithResolvedUrls.push({
+                label: 'skipReason',
+                content: reason
+            });
+        } else if (!Array.isArray(reason) && typeof reason.props.children === 'string' && typeof reason.props.href === 'string') {
```
DudaGod, line 121: Reason can be an array?

lib/static/new-ui/features/suites/components/TreeViewItemSubtitle/index.tsx
```diff
@@ -16,7 +17,11 @@ interface TreeViewItemSubtitleProps {
 }
 
 export function TreeViewItemSubtitle(props: TreeViewItemSubtitleProps): ReactNode {
-    if (props.item.images?.length) {
+    if (props.item.skipReason) {
+        return <div className={styles.skipReasonContainer}>
+            <div className={styles.skipReason}>Skipped ⋅ {Parser(props.item.skipReason)}</div>
```
DudaGod, line 22: ` ⋅ ` - why you choose a dot as a separator?  But in IU it looks quite harmonious. Maybe you use it in some other components?

DudaGod (approved):
🔥

lib/static/new-ui/features/suites/components/TreeViewItemSubtitle/index.tsx
shadowusr: I think it looks nice as a separator and it's not the only place we use dot. The other places are hint toast, and other variants of subtitle in tree view.

lib/static/new-ui/components/MetaInfo/index.tsx
shadowusr: We discussed this with sipayrt and decided that it's not very good idea to include such specific logic related to `muteReason`, so i got rid of altogether

lib/static/new-ui/components/MetaInfo/index.tsx
shadowusr: This logic is no longer here.

lib/static/new-ui/components/MetaInfo/index.tsx
shadowusr: This is no longer here, got rid of parsing html entirely.

PR #627, opened 2025-01-03
feat: new ui usage information tracking

shadowusr:
### What's done?
- Added a sign of new ui in visit params
- Added tracking of all major new-ui features (sorting, grouping, toolbar, redux actions and more)
- Enabled ya metrika by default for all users and added its script
- Implemented e2e tests for analytics info collection
- Updated docs

### DIFF
--- a/lib/config/index.ts
+++ b/lib/config/index.ts
@@ -28,7 +28,7 @@ const assertType = <T>(name: string, validationFn: (value: unknown) => value is 
 };
 const assertString = (name: string): AssertionFn<string> => assertType(name, _.isString, 'string');
 const assertBoolean = (name: string): AssertionFn<boolean> => assertType(name, _.isBoolean, 'boolean');
-const assertNumber = (name: string): AssertionFn<number> => assertType(name, _.isNumber, 'number');
+export const assertNumber = (name: string): AssertionFn<number> => assertType(name, _.isNumber, 'number');
 const assertPlainObject = (name: string): AssertionFn<Record<string, unknown>> => assertType(name, isPlainObject, 'plain object');
 
 const assertSaveFormat = (saveFormat: unknown): asserts saveFormat is SaveFormat => {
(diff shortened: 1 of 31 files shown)


lib/config/index.ts
```diff
@@ -206,11 +206,27 @@ const getParser = (): ReturnType<typeof root<ReporterConfig>> => {
...
+                map: (value: boolean) => {
+                    if (process.env.NO_ANALYTICS && JSON.parse(process.env.NO_ANALYTICS)) {
+                        return false;
+                    }
+
+                    return value;
+                }
+            }),
             counterNumber: option({
+                isDeprecated: true,
```
KuznetsovRoman, line 225: For this it is necessarily to update gemini-configparser.
Currently it is 1.0.0 in html-reporter and latest is 1.4.2.
1.4.2 - fix for "isDeprecated"

test/func/tests/analytics/index.testplane.js
```diff
@@ -0,0 +1,117 @@
...
+
+        await browser.$('[data-qa="footer-item-info"]').click();
+
+        const [metrikaCall] = await browser.execute(() => {
+            return window.ym.calls;
+        });
+
+        expect(metrikaCall[1]).toEqual('reachGoal');
+        expect(metrikaCall[2]).toEqual('FEATURE_USAGE');
+        expect(metrikaCall[3]).toEqual({featureName: 'Open info panel'});
```
KuznetsovRoman, line 73: not critical: This check looks fragile and may break if this click would produce other metrika action before this one. Maybe we should save all `metricaCalls` and check if it has any call, matching args `['reachGoal', 'FEATURE_USAGE', {featureName: 'Open info panel'}]`?

KuznetsovRoman (approved):
Need to check "what data actually is being sent with analytics available" to be sure we are not sending anything extra.

Other than that, /ok

PR #634, opened 2025-02-04
fix: fix diff mode select

Hypnosphi:
#626 unfortunately broke the diff mode select, this fixes it


Hypnosphi:
@shadowusr @KuznetsovRoman can you please take a look

KuznetsovRoman:
Thanks for the contribution!

PR #640, opened 2025-03-18
chore: use github ci

KuznetsovRoman:


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,219 @@
...
+    - name: Deploy Testplane html-reporter reports
+      if: ${{ env.IS_FAILED }}
+      uses: peaceiris/actions-gh-pages@v4
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        publish_dir: ${{ env.SOURCE_REPORTS_DIR }}
+        destination_dir: ${{ env.DEST_REPORTS_DIR }}
+        keep_files: true
+
+    - name: Construct failed PR comment
```
KuznetsovRoman, line 145: Couldn't figure out smarter way to do this.
I tried constructing variables like `STEP_TESTPLANE_TESTPLANE_COMMON_OUTCOME` and evaluating them, but it does not work

test/func/fixtures/playwright/tests/failed-describe.spec.ts
```diff
@@ -38,7 +38,7 @@ test.describe('failed describe', () => {
     test('test with successful assertView and error', async ({page, baseURL}) => {
         await page.goto(baseURL as string);
 
-        await expect(page.locator('header')).toHaveScreenshot('header-success.png');
+        await expect(page.locator('header')).toHaveScreenshot('header-success.png', {maxDiffPixelRatio: 1});
```
KuznetsovRoman, line 41: As its "test with successful assertView" (fixture), we can make this test 100% stable

test/func/fixtures/plugins/success-describe.testplane.js
```diff
@@ -8,6 +8,6 @@ describe('success describe', function() {
     it('test with screenshot', async ({browser}) => {
         await browser.url(browser.options.baseUrl);
 
-        await browser.assertView('header', 'header');
+        await browser.assertView('header', 'header', {ignoreDiffPixelCount: '100%'});
```
KuznetsovRoman, line 11: As its "success-describe" (fixture), we can make this test 100% stable

test/func/fixtures/testplane-tinder/.testplane.conf.js
KuznetsovRoman, line 1: Copied from "testplane-gui"

test/func/packages/html-reporter-test-server/index.js
```diff
@@ -18,17 +18,24 @@ module.exports = (testplane, pluginConfig) => {
...
                 throw new Error(err);
             }
 
             console.info(`Server is listening on http://localhost:${port}`);
+        }).on('error', err => {
+            if (err.code === 'EADDRINUSE') {
+                console.warn(`Skip running server, because port ${port} is busy`);
+            } else {
+                throw err;
+            }
```
KuznetsovRoman, line 34: - We run Testplane (which uses this html-reporter-test-server and opens the server on SERVER_PORT)
  - Testplane runs Testplane GUI (port GUI_PORT)
    - Testplane GUI uses html-reporter-test-server and tries to run server on SERVER_PORT

test/func/tests/common/test-results-appearance.testplane.js
```diff
@@ -85,12 +85,13 @@ describe('Test results appearance', () => {
...
+                getTestStateByNameSelector('header') + '/..'
+            );
+
             for (const field of ['message', 'name', 'stack']) {
-                const errorMessage = await browser.$(
-                    getTestSectionByNameSelector('test without screenshot') +
-                    getTestStateByNameSelector('header') +
-                    getElementWithTextSelector('span', field) + '/..'
-                );
+                const errorMessage = await parentElement.$(getElementWithTextSelector('span', field) + '/..');
```
KuznetsovRoman, line 94: For some reason, it didn't work that way (though only for "name")

test/func/tests/plugins/tests-redux-plugin.testplane.js
```diff
@@ -15,7 +15,7 @@ describe('Test redux plugin', function() {
         await browser.$(screenSelector).waitForDisplayed();
         await browser.$(clickSelector).click();
         // Letting browser handle re-renders
-        await browser.pause(300);
+        await browser.pause(1000);
```
KuznetsovRoman, line 18: 500 wasn't enough either

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,219 @@
...
+          comment+="- ❌ testplane-tinder: failure. "
+        fi
+        comment+="[Report](${link_base}/testplane-tinder) <br>"
+
+        echo "FAILED_PR_COMMENT=${comment}" >> $GITHUB_ENV
+
+    - name: Leave failed comment to PR with link to Testplane HTML reports
+      if: env.IS_FAILED
+      id: leave-failed-comment-step
+      uses: thollander/actions-comment-pull-request@v3
```
KuznetsovRoman, line 205: Example: 
Screenshot 2025-03-20 at 02 28 22

shadowusr (approved):
I'm ok with the current implementation, because it fixes issues with CircleCI & is convenient to use from the user perspective.

What I don't like about this new approach is that workflow file is pretty complex and most importantly if we add new test project we should not forget to add it to this workflow, which is pretty unexpected.

I think the code could be shortened if loops were used:
```
    - name: Run all Testplane test projects
      id: run-tests
      working-directory: 'test/func/tests'
      run: |
        echo "FAILED_TESTS=" >> $GITHUB_ENV
        echo "TEST_RESULTS=" >> $GITHUB_ENV
        
        IFS=',' read -ra TEST_ARRAY <<< "$TEST_PROJECTS"
        for TEST in "${TEST_ARRAY[@]}"; do
          echo "Running test project: $TEST"
          if npm run testplane:$TEST; then
            echo "✅ $TEST test passed"
            echo "TEST_RESULTS=$TEST_RESULTS- ✅ $TEST: success. [Report](\${link_base}/$TEST) <br>" >> $GITHUB_ENV
          else
            echo "❌ $TEST test failed"
            echo "FAILED_TESTS=$FAILED_TESTS $TEST" >> $GITHUB_ENV
            echo "TEST_RESULTS=$TEST_RESULTS- ❌ $TEST: failure. [Report](\${link_base}/$TEST) <br>" >> $GITHUB_ENV
          fi
        done
        
        if [ ! -z "$FAILED_TESTS" ]; then
          echo "IS_FAILED=true" >> $GITHUB_ENV
        fi
```

But this is still pretty complex & requires us to hard-code test projects in the beginning of the file.

Another approach would be to merge all those reports into one, using our merge-reports command. It would be both easy to write this workflow & convenient to work with from the user's perspective. What do you think?

But if that's too much of a hassle feel free to merge it as is for now...

PR #646, opened 2025-04-02
fix: html-reporter gui playwright test launch (step.startTime.getTime)

KuznetsovRoman:


lib/adapters/test-result/playwright.ts
```diff
@@ -159,7 +159,7 @@ const getHistory = (steps?: PlaywrightTestResult['steps']): TestStepCompressed[]
         [TestStepKey.Name]: step.title,
         [TestStepKey.Args]: [],
         [TestStepKey.IsFailed]: Boolean(step.error),
-        [TestStepKey.TimeStart]: step.startTime.getTime(),
+        [TestStepKey.TimeStart]: step.startTime instanceof Date ? step.startTime.getTime() : new Date(step.startTime).getTime(),
```
KuznetsovRoman, line 162: When launched via "npx playwright test", its intance of Date.
When launched via "npx html-reporter --tool playwright", its ISO string

test/unit/lib/adapters/test-result/playwright.ts
```diff
@@ -168,6 +168,27 @@ describe('PlaywrightTestResultAdapter', () => {
...
+                [TestStepKey.Children]: []
+            }), mkTestStepCompressed({
+                [TestStepKey.Name]: 'Step2',
+                [TestStepKey.Duration]: 200,
+                [TestStepKey.TimeStart]: 2000,
+                [TestStepKey.Children]: []
+            })];
+
+            assert.deepEqual(adapter.history, expectedHistory);
+        });
```
DudaGod, line 191: the test differs from the previous one only by the field `startTime`. It seems we can get rid of duplication by using `forEach`

DudaGod (approved):

test/unit/lib/adapters/test-result/playwright.ts
KuznetsovRoman: as those are just 2 tests, i think its more readable that way

PR #647, opened 2025-04-04
fix(merge-reports): ability to merge reports locally

DudaGod:
### What is done

Previously, merge reports only worked with urls on `databaseUrls.json` and `sqlite.db`. If user send path to these files on file system merge was complete successfully but in merged report the tests were not displayed.

So in this PR I fix this behaviour and now merge reports correctly work with files on fs.

What can be improved: ability to path just a folder.

### DIFF
--- a/lib/merge-reports/index.js
+++ b/lib/merge-reports/index.js
@@ -1,12 +1,19 @@
 'use strict';
 
+const url = require('url');
+const urljoin = require('url-join');
+const path = require('path');
+const fs = require('fs-extra');
 const axios = require('axios');
 const _ = require('lodash');
+
 const serverUtils = require('../server-utils');
+const {isUrl, logger} = require('../common-utils');
 const dbServerUtils = require('../db-utils/server');
+const {DB_FILE_EXTENSION, IMAGES_PATH, SNAPSHOTS_PATH, ERROR_DETAILS_PATH, DATABASE_URLS_JSON_NAME, LOCAL_DATABASE_NAME} = require('../constants');
 
 module.exports = async (toolAdapter, srcPaths, {destPath, headers}) => {
-    validateOpts({srcPaths, destPath, headers});
+    await validateOpts({srcPaths, destPath, headers});
 
     let headersFromEnv;
     const {htmlReporter, reporterConfig} = toolAdapter;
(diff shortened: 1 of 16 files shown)


lib/merge-reports/index.js
```diff
@@ -90,3 +119,25 @@ async function tryResolveUrl(url, headers) {
...
+function genUniqDbName(dbPath, num) {
+    return `${path.basename(dbPath, DB_FILE_EXTENSION)}_${num}${DB_FILE_EXTENSION}`;
+}
+
+async function copyDbFiles(dbPaths) {
+    await Promise.all(dbPaths.map(({src, dest}) => fs.copy(src, dest)));
+}
+
+async function copyArtifacts({srcPaths, destPath, folderName}) {
+    for (const reportPath of srcPaths) {
```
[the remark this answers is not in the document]
KuznetsovRoman: You've got race condition, while trying to create the same directory multiple times at the same time. It happens under the hood of `"fs-extra".copy`

Here is an easy fix to get 80% of result by 20% of time spent:
- Find uniq destination directories
- Create them one by one (as you probably wouldn't want to write a lot of code to "create first-level directories first", "then second-level directories", ...). It does not look like a problem as "creating a directory" should be fast
- Copy all the files with `await Promise.all` without `recursive: true`

As you mentioned "fs.copy" is slow, i think its important to perform it in parallel

KuznetsovRoman (approved):
I think "parallel fs.copy" is important, though its fine if you would like to fix it in another PR

shadowusr (changes requested):
~~Tried to switch to this branch & test it locally, stumbled upon a few issues ...~~

shadowusr (approved):
Issues I've encountered:
- we need an issue to fix merge-reports for playwright reports
- we need to fix or create an issue to fix merging reports that aren't local and their databases don't have absolute URLs to images/snapshots. What happens in that case: we create a databaseUrls.json with links to those databases. Those databases still have relative urls to images, etc and it won't work
- I think it's a good idea to add info to help that databaseUrls.json need to be specified, not just report directories

lib/merge-reports/index.js
```diff
@@ -24,24 +31,81 @@ module.exports = async (toolAdapter, srcPaths, {destPath, headers}) => {
...
 
+    if (srcPaths.length === 1) {
+        throw new Error(`Nothing to merge, only one source report is passed: ${srcPaths[0]}`);
+    }
+
     if (srcPaths.includes(destPath)) {
         throw new Error(`Destination report path: ${destPath}, exists in source report paths`);
     }
 
+    for (const srcPath of srcPaths) {
```
DudaGod, line 78: here I've added more different cases that need to be validated here

lib/merge-reports/index.js
```diff
@@ -51,7 +115,15 @@ function validateOpts({srcPaths, destPath, headers}) {
 
 async function tryResolveUrls(urls, headers) {
     const resolvedUrls = [];
-    const results = await Promise.all(urls.map(url => tryResolveUrl(url, headers)));
+    const results = await Promise.all(urls.map(async u => {
+        const extName = path.extname(isUrl(u) ? url.parse(u).pathname : u);
+
+        if (!extName) {
+            u = isUrl(u) ? urljoin(u, DATABASE_URLS_JSON_NAME) : path.join(u, DATABASE_URLS_JSON_NAME);
```
DudaGod, line 122: Here support case if user send folder on html report

PR #648, opened 2025-04-06
chore(ci): setup pkg.pr.new

DudaGod:


.github/workflows/node.js.yml
```diff
@@ -23,3 +23,7 @@ jobs:
         cache: 'npm'
     - run: npm ci
     - run: npm test
+
+    - name: Publish
+      if: ${{ matrix.node-version == '20.x' }}
```
DudaGod, line 28: There is no need to publish package from few node versions

KuznetsovRoman (approved):

PR #652, opened 2025-04-17
fix: do not require testplane when use report with other tools

DudaGod:


lib/gui/server.ts
```diff
@@ -14,11 +14,6 @@ import {ServerReadyData} from './api';
 import {BrowserFeature, ToolName} from '../constants';
 import type {TestplaneToolAdapter} from '../adapters/tool/testplane';
 import {ToolRunnerTree} from './tool-runner';
-import {
-    getSnapshotHashWithoutAttempt,
-    snapshotsInProgress,
-    TestContext
-} from '../adapters/event-handling/testplane/snapshots';
```
DudaGod, line 21: Testplane is connected inside this module, but it doesn't exist in project with playwright for example. So I use `await import` below in the code only for testplane tool

shadowusr (approved):

lib/adapters/event-handling/testplane/snapshots.ts
```diff
@@ -87,7 +87,7 @@ export const finalizeSnapshotsForTest = async ({testResult, attempt, reportPath,
             return [];
         }
 
-        const shouldSave = recordConfig.mode !== RecordMode.LastFailedRun || (eventName === events.TEST_FAIL);
+        const shouldSave = RecordMode && (recordConfig.mode !== RecordMode.LastFailedRun || (eventName === events.TEST_FAIL));
```
DudaGod, line 90: Support case when `RecordMode` is not imported from testplane (old versions)

shadowusr (approved):

PR #654, opened 2025-04-18
fix: static accepter image load fail

KuznetsovRoman:


lib/static/modules/actions/static-accepter.ts
```diff
@@ -83,13 +84,19 @@ export const staticAccepterCommitScreenshot = (
...
-                const blob = await getBlob(imageInfo.image);
+            for (let i = 0; i < imagesInfo.length; i++) {
+                queue.add(async () => {
+                    const blob = await getBlobWithRetires(imagesInfo[i].image);
 
-                payload.append('image', blob, imageInfo.path);
-            }));
+                    payload.append('image', blob, imagesInfo[i].path);
+
+                    dispatch({type: actionNames.UPDATE_LOADING_TITLE, payload: `Preparing images to commit: ${i + 1} of ${imagesInfo.length}`});
```
KuznetsovRoman, line 95: dispatching this action *after* the blob is downloaded seems more suitable.

### Example: we have 256 images.

#### Before:
We instantly dispatch 256 actions and user sees "preparing to commit: 256/256" (and only after we only start downloading these images)

#### After:
We update the counter after download is complete. So at start user will see "0/256" while nothing was downloaded

lib/static/modules/utils/index.js
```diff
@@ -210,17 +210,34 @@ export function parseKeyToGroupTestsBy(key) {
     return [groupSection, groupKey];
 }
 
-export function getBlob(url) {
-    return new Promise((resolve, reject) => {
-        const xhr = new XMLHttpRequest();
+export async function getBlobWithRetires(url, retriesCount = 3) {
```
KuznetsovRoman, line 213: We can also be rps limited
So exponential retries (1s, 2s, 4s) would not be superfluous

lib/static/modules/actions/static-accepter.ts
```diff
@@ -83,13 +84,19 @@ export const staticAccepterCommitScreenshot = (
                 payload.append('meta', JSON.stringify(meta));
             }
 
-            await Promise.all(imagesInfo.map(async (imageInfo, index) => {
-                dispatch({type: actionNames.UPDATE_LOADING_TITLE, payload: `Preparing images to commit: ${index + 1} of ${imagesInfo.length}`});
+            const queue = new PQueue({concurrency: 256});
```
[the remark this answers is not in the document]
shadowusr: if this works then ok, but if 256 is the max value in most browsers, maybe it's worth limiting it slightly below the limit to preserve room for other, unrelated to images uploading requests? For example, so that when user is uploading images, he could still use report (including downloading time travel snapshots or downloading screenshots for other tests).

shadowusr (approved):
Awesome enhancement! 🔥

DudaGod (approved):

PR #658, opened 2025-05-03
fix: implement backwards compatibility with old sqlite formats

shadowusr:
### What's done?
- Implement sqlite dbs versioning via separate version table
- v0 is sqlite db format before introduction of time travel
- From now on, all future html-reporter versions should be backwards compatible with v0, and there's e2e test covering that

When user opens v0 db and upwards in GUI mode, an automatic migration is applied, all data is retains and user seamlessly sees UI without any interruptions.

When user tries to open db of unsupported version (older than v0 or newer than current html-reporter version), this unsupported report is moved to a new directory, a warning is printed to console and user sees fresh UI (without any previous test results).


test/func/fixtures/db-migrations/report/index.html
shadowusr, line 1: Committed the whole report directory to freeze the old, v0 report forever. This old v0 report is used to e2e check migrations logic of backwards-compatibility.

lib/sqlite-client.ts
```diff
@@ -90,7 +91,12 @@ export class SqliteClient {
             debug('db connection opened');
 
             createTablesQuery().forEach((query) => db?.prepare(query).run());
+
+            if (!options.reuse) {
+                setDatabaseVersion(db, DB_CURRENT_VERSION);
+            }
         } catch (err: any) { // eslint-disable-line @typescript-eslint/no-explicit-any
+            console.dir(err);
```
KuznetsovRoman, line 99: This console.dir looks extra

test/func/fixtures/db-migrations/failed-describe.testplane.js
```diff
@@ -0,0 +1,39 @@
...
+    });
+
+    it('test with long error message', async () => {
+        throw new Error(`long_error_message ${'0123456789'.repeat(20)}\n message content`);
+    });
+
+    it('test with successful assertView and error', async ({browser}) => {
+        await browser.url(browser.options.baseUrl);
+
+        await browser.assertView('header', 'header');
```
KuznetsovRoman, line 27: If it is "successful assertView" lets add `{ignoreDiffPixelCount: '100%'}` as 3rd argument to make it 100% stable

test/func/fixtures/db-migrations/success-describe.testplane.js
```diff
@@ -0,0 +1,13 @@
...
+    it('successfully passed test', async ({browser}) => {
+        await browser.url(browser.options.baseUrl);
+
+        assert.isTrue(true);
+    });
+
+    it('test with screenshot', async ({browser}) => {
+        await browser.url(browser.options.baseUrl);
+
+        await browser.assertView('header', 'header');
```
KuznetsovRoman, line 11: the same: `{ignoreDiffPixelCount: '100%'}` would not hurt

test/func/tests/db-migrations/index.testplane.js
```diff
@@ -0,0 +1,37 @@
...
+    afterEach(async () => {
+        await treeKill(guiProcess.pid);
+
+        childProcess.execSync('git checkout .', {cwd: reportDir});
+    });
+
+    // It's sufficient to check only v0, because that means all newer versions are also supported
+    // through a chain of migrations v0 -> v1 -> ...
+    it('should be able to open report with sqlite.db v0', async ({browser}) => {
+        await browser.pause(2000);
```
KuznetsovRoman, line 33: Why is there 2s pause at the start of the test?
Maybe we should move this pause to `beforeEach`?

test/unit/lib/gui/tool-runner/utils.ts
```diff
@@ -0,0 +1,165 @@
...
+        afterEach(async () => {
+            consoleWarnStub.restore();
+            await fs.remove(tempDir);
+        });
+
+        it('should do nothing if database file does not exist', async () => {
+            await utilsWithStubs.prepareLocalDatabase(reportPath);
+
+            sinon.assert.notCalled(backupAndResetStub);
+            sinon.assert.notCalled(consoleWarnStub);
```
KuznetsovRoman, line 72: We usually call assert from global scope

KuznetsovRoman (approved):
I would add "ignoreDiffPixelCount" to new fixtures of successful screenshot examples.

Other than that, seems great!

lib/db-utils/migrations.ts
```diff
@@ -0,0 +1,114 @@
...
+import {createTableQuery} from './common';
+const debug = makeDebug('html-reporter:db-migrations');
+
+export const getDatabaseVersion = (db: Database.Database): number | null => {
+    debug('getDatabaseVersion');
+    try {
+        const versionTableExists = db.prepare(
+            `SELECT name FROM sqlite_master WHERE type='table' AND name=?`
+        ).get(DB_VERSION_TABLE_NAME);
+        debug('versionTableExists: ', versionTableExists);
```
DudaGod, line 22: It seems that you don't have to put a space here and below. In this case two spaces will appear in the log

lib/db-utils/migrations.ts
```diff
@@ -0,0 +1,114 @@
...
+            if (!isNaN(version) && version >= 0) {
+                return version;
+            }
+        }
+
+        const tableInfo = db.prepare(`PRAGMA table_info(${DB_SUITES_TABLE_NAME})`).all() as {name: string}[];
+        const columnNames = tableInfo.map(col => col.name);
+        debug('column names in db: ', columnNames);
+
+        const version0Columns = [
```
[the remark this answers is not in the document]
shadowusr: I agree it's not the prettiest chunk of code, but IMO it's the only way. We can't just say "if there's no version table, it's v0". We guarantee backwards compatibility with all versions, starting with v0. And to satisfy that guarantee, we should implement migrations from v0. Those migrations can only work with very strict db structure, having exactly those columns. If it doesn't match, applying migration would just corrupt db.

What if db has no version table and doesn't have duration column? In that case it's better to just create a backup and start from scratch instead of incorrectly applying migration.

Therefore I'm keeping it for now.

test/func/fixtures/db-migrations/failed-describe.testplane.js
shadowusr: These are never intended to run, report should be fixed, but just in case, added that option.

test/func/tests/db-migrations/index.testplane.js
shadowusr: Reduced delay and use `screenshotDelay` instead. It's needed because, some elements don't get rendered in time.

PR #661, opened 2025-05-07
fix: do not throw when trying to merge single report

shadowusr:


KuznetsovRoman (approved):

lib/merge-reports/index.js
```diff
@@ -68,7 +68,7 @@ async function validateOpts({srcPaths, destPath, headers}) {
     }
 
     if (srcPaths.length === 1) {
-        throw new Error(`Nothing to merge, only one source report is passed: ${srcPaths[0]}`);
+        console.warn(`Nothing to merge, only one source report is passed: ${srcPaths[0]}`);
```
DudaGod, line 71: Looks like we should change message here, because in this case merged report will be created

DudaGod (approved):

PR #664, opened 2025-05-14
fix: improve cross-browser compatibility and update gravity-ui

shadowusr:
### What's done?

- Updated Gravity UI to the latest version and related packages (visually this resulted in a change of filtering by status)
- Report is now working correctly in all major browsers (tested on chrome, firefox, safari)
- Minor infrastructure/tsconfig updates

DEMO report can be viewed via the link in the comment by robot below!


test/setup/globals.js
```diff
@@ -7,6 +7,9 @@ Promise.config({longStackTraces: true});
 global.sinon = require('sinon');
 global.assert = chai.assert;
 
+global.requestAnimationFrame = () => {};
+global.cancelAnimationFrame = () => {};
```
KuznetsovRoman, line 11: Since we have sinon here, maybe these should be sinon stubs?

KuznetsovRoman (approved):

test/setup/globals.js
shadowusr: Honestly don't see why this can be useful. Every stub will store full call history, but we never test requestAnimationFrames or anything like that so that would just be unnecessary. I needed to add these because after update gravity-ui started using floating-ui, which uses requestAnimationFrame and caused tests to fail since jsdom doesn't implement these.

I think it would be better to keep it as is for now and if we need to test animations, stub these on a per-test basis.

PR #666, opened 2025-05-19
feat: implement player show/hide button

shadowusr:
### What's done?

- Fixed visuals when errored step is highlighted
- Added a button to hide Time Travel Player
- Implemented ability to change playback speed
- Added config overrides to enable Time Travel by default for newer versions of Testplane

See demo report via robot link below.


lib/static/new-ui/features/suites/components/TestControlPanel/index.tsx
```diff
@@ -49,6 +54,15 @@ export function TestControlPanel(props: TestControlPanelProps): ReactNode {
         }
     };
 
+    const onTogglePlayerVisibility = (): void => {
+        analytics?.trackFeatureUsage({featureName: 'Toggle time travel player visibility'});
+        dispatch(toggleTimeTravelPlayerVisibility(!isPlayerVisible));
+    };
+
+    const showRetryButton = !!isRunTestsAvailable;
```
DudaGod, line 62: Why not just `Boolean(...)`?

lib/static/new-ui/features/suites/components/SnapshotsPlayer/index.tsx
```diff
@@ -12,16 +12,33 @@ import {getCurrentResult} from '@/static/new-ui/features/suites/selectors';
...
+        label: 'Speed',
+        data: {
+            hint: <span><b style={{fontWeight: 600}}>Note:</b> transitions or animations will still play at original speed.</span>
+        },
+        options: [
+            {value: '0.5', content: '0.5x'},
+            {value: '1', content: '1x'},
+            {value: '1.5', content: '1.5x'},
+            {value: '2', content: '2x'},
+            {value: '4', content: '4x'}
```
DudaGod, line 38: to my mind we can stop at 2x (like in most services, youtube for example). If the test lasts longer than a minute then we can just move to the desired part of the test

lib/static/new-ui/features/suites/components/SnapshotsPlayer/index.tsx
```diff
@@ -251,6 +269,48 @@ export function SnapshotsPlayer(): ReactNode {
         }
     }, []);
 
+    const onSpeedChange = useCallback((value: string[]): void => {
+        console.log(value);
```
DudaGod, line 273: looks like wa can remove it

lib/static/new-ui/features/suites/components/SnapshotsPlayer/index.tsx
```diff
@@ -251,6 +269,48 @@ export function SnapshotsPlayer(): ReactNode {
...
+        if (value) {
```
DudaGod, line 274: Here we can simplify code a little bit:
```
if (!value) {
  return;
}

// ...
```

Allows you to reduce the amount of nested code

lib/gui/config-overrides.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const debug = makeDebug('html-reporter:testplane');
+
+export const overrideTestplaneConfig = (toolName: ToolName, config: TestplaneConfigAdapter): void => {
+    if (toolName !== ToolName.Testplane) {
+        debug(`Skipping config overrides because toolName is not Testplane`);
+        return;
+    }
+
+    if (process.env.HTML_REPORTER_NO_CONFIG_OVERRIDES) {
```
DudaGod, line 15: will we write about this environment variable somewhere in the documentation?

lib/gui/index.ts
```diff
@@ -27,6 +29,8 @@ export interface ServerArgs {
 }
 
 export default (args: ServerArgs): void => {
+    overrideTestplaneConfig(args.toolAdapter.toolName, args.toolAdapter.config as TestplaneConfigAdapter);
```
DudaGod, line 32: I'm not sure that we should override user config in gui mode. Because to my mind user must set the required values himself. I would just change the default value of option.

lib/static/new-ui/features/suites/components/SnapshotsPlayer/index.tsx
shadowusr: I think it won't harm to have it here tbh. I myself often need more than 2x on youtube

lib/gui/config-overrides.ts
shadowusr: Discussed & decided to implement this ability in the UI instead of using env vars over here.

lib/gui/index.ts
shadowusr: Decided to do it using a toggle in Settings panel in UI.

PR #671, opened 2025-05-28
fix: do not fail if test statistics is empty

DudaGod:


lib/static/modules/reducers/stats.js
```diff
@@ -4,7 +4,7 @@ export default (state, action) => {
     switch (action.type) {
         case actionNames.INIT_STATIC_REPORT: {
             const {stats} = action.payload;
-            const {perBrowser, ...restStats} = stats;
+            const {perBrowser, ...restStats} = stats || {};
```
DudaGod, line 7: When info about tests is not loaded we send payload with `stats: null` - https://github.com/gemini-testing/html-reporter/blob/master/lib/static/modules/actions/lifecycle.ts#L127

Fixed it here because I'm not sure that someone not use this field with `null` value in some plugin for html-reporter

shadowusr: looks like this file is really tiny and if this was in typescript, the error wouldn't have happened. Can we re-write it in typescript?

All actions are now strictly typed.

shadowusr (approved):

PR #674, opened 2025-06-10
fix: create destination directory beforehand during merge-reports

shadowusr:


lib/report-builder/static.ts
```diff
@@ -65,6 +65,7 @@ export class StaticReportBuilder {
     async saveStaticFiles(): Promise<void> {
         const destPath = this._reporterConfig.path;
 
+        await fs.ensureDir(destPath);
```
shadowusr, line 68: Same fix as above.

PR #676, opened 2025-06-12
fix: fix scroll issues in old ui

shadowusr:
### What's done?

This PR fixes scroll issues in old ui, by making diff viewer size more deterministic. The issue was that in certain cases when image is not loaded, size wasn't computed correctly.

Demo reports coming soon.


sipayRT (approved):
not for this PR: we need a common css-file for the DiffViewer component to avoid css duplication

PR #689, opened 2025-07-24
feat(new-ui): squash repeated steps

sonic16x:
Screenshot 2025-07-25 at 14 07 06


lib/adapters/test-result/testplane.ts
```diff
@@ -52,25 +55,161 @@ const wrapSkipComment = (skipComment: string | null | undefined): string => {
...
-            result[TestStepKey.Children] = getHistory(h[TestStepKey.Children]);
+const testItemsTheSame = (a: TestStepCompressed, b: TestStepCompressed): boolean => (
+    a &&
+    (!a[TestStepKey.Children] || !b[TestStepKey.Children]) &&
+    (!a[TestStepKey.IsFailed] || !b[TestStepKey.IsFailed]) &&
+    a[TestStepKey.Name] === b[TestStepKey.Name] &&
+    a[TestStepKey.Args].join() === b[TestStepKey.Args].join()
+);
+
+const arraysEqual = <I>(a: I[], b: I[], checkFunc: (a: I, b: I) => boolean): boolean => {
```
shadowusr, line 66: I personally don't like such very generic and "universal" helpers, imho they don't bring much value beyond core APIs, and are often forgotten about the minute this PR is merged. FWIW, we already have lodash as a dependency and if you don't want to compare length twice so much, you could use `_.isEqualWith`

But that's just my personal opinion, you don't have to fix this. It's a matter of taste.

lib/adapters/test-result/testplane.ts
```diff
@@ -52,25 +55,161 @@ const wrapSkipComment = (skipComment: string | null | undefined): string => {
...
+    for (let i = 0; i < (repeat - 1); i++) {
+        total += items[start + (i * groupLen) + index][TestStepKey.Duration];
+    }
+
+    return parseFloat((total / (repeat - 1)).toFixed(2));
+};
+
+const MIN_REPEATS = 3; // Min count of repeats elements of group elements for squash
+
+const collapseRepeatingGroups = (
```
shadowusr, line 102: I think we need unit tests on this

shadowusr (approved):
I think this looks absolutely great! 🔥 

Before merging, I think it would be cool to check how sqlite db size changes with this new logic and implement unit tests on new helpers

lib/adapters/test-result/testplane.ts
sonic16x: It is better than is _. isEqualWith because check also length

lib/adapters/test-result/testplane.ts
```diff
@@ -52,25 +55,161 @@ const wrapSkipComment = (skipComment: string | null | undefined): string => {
...
-            [TestStepKey.IsFailed]: h[TestStepKey.IsFailed],
-            [TestStepKey.IsGroup]: h[TestStepKey.IsGroup]
-        };
-
-        if (h[TestStepKey.Children] && h[TestStepKey.IsFailed]) {
-            result[TestStepKey.Children] = getHistory(h[TestStepKey.Children]);
+const testItemsTheSame = (a: TestStepCompressed, b: TestStepCompressed): boolean => (
+    a &&
+    (!a[TestStepKey.Children] || !b[TestStepKey.Children]) &&
+    (!a[TestStepKey.IsFailed] || !b[TestStepKey.IsFailed]) &&
```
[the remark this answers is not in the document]
sonic16x: We don't collapse items if they failed

lib/adapters/test-result/testplane.ts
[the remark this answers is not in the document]
sonic16x: I add results with sizes to task

PR #700, opened 2025-08-21
chore!: remove compiled dependencies

KuznetsovRoman:
Changes Public API:
Event [DATABASE_CREATED](https://github.com/gemini-testing/html-reporter/blob/master/docs/en/html-reporter-events.md#database_created) will now pass "sql.js" database instead of "better-sqlite3"


lib/db-utils/client.js
```diff
@@ -92,7 +92,7 @@ export function getSuitesTableRows(db) {
 }
 
 export function getTableRows(db, tableName) {
-    const [databaseRows] = db.exec(selectAllQuery(tableName));
+    const databaseRows = db.exec(selectAllQuery(tableName))[0];
```
KuznetsovRoman, line 95: everywhere "[0]" at the end of "exec" means "first executed statement" because "sql.js" supports executing multiple js statements in "exec"

shadowusr (approved):
Looking great! 🔥 

Still a bit hesitant about publishing next major version without further considerations... But it's fine if no one can think of other breaking changes we'd want to pack into v11.

Also, we probably need to update events docs if they are the reason for major version, don't we?

PR #701, opened 2025-08-22
chore: optimize package size

KuznetsovRoman:
- Splitted webpack chunks, extracting common code, which reduced build time (by 4.5 seconds) and bundles size:

<table>
  <tr>
    <td align="center"><b>Before</b></td>
    <td align="center"><b>After</b></td>
  </tr>
  <tr>
    <td>
      before
    </td>
    <td>
      <table>
        <tr>
          <td>
image
          </td>
        </tr>
      </table>
    </td>
  </tr>
</table>

image

- Removed `@babel/runtime` out from prod deps

- Replaced old "inquirer" with new light "@inquirer/confirm"

### DIFF
--- a/lib/cli/commands/remove-unused-screens/index.js
+++ b/lib/cli/commands/remove-unused-screens/index.js
@@ -9,7 +9,7 @@ const filesize = require('filesize');
 const Promise = require('bluebird');
 
 const {commands} = require('../..');
-const {getTestsFromFs, findScreens, askQuestion, identifyOutdatedScreens, identifyUnusedScreens, removeScreens} = require('./utils');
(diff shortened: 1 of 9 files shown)


shadowusr (approved):
Looks awesome to me, such a drastic reduction in package weight!

One thing that's a bit concerning to me is the ease of debugging... I bet with all these chunks it's almost impossible to find out where the error's coming from? Does it work fine in dev mode?

KuznetsovRoman:
> Looks awesome to me, such a drastic reduction in package weight!
> 
> One thing that's a bit concerning to me is the ease of debugging... I bet with all these chunks it's almost impossible to find out where the error's coming from? Does it work fine in dev mode?

With dev build, there does not seem to be any difference. Actual js looks terrible, but source maps work just fine:
image

PR #705, opened 2025-09-02
perf: speed up tests tree builder construction

KuznetsovRoman:
We used to have 2 places with square asymptotics, and now they are linear as well

### DIFF
--- a/lib/tests-tree-builder/base.ts
+++ b/lib/tests-tree-builder/base.ts
@@ -1,4 +1,3 @@
-import _ from 'lodash';
 import {determineFinalStatus} from '../common-utils';
 import {BrowserVersions, DEFAULT_TITLE_DELIMITER, TestStatus} from '../constants';
 import {ReporterTestResult} from '../adapters/test-result';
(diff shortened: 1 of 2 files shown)


lib/tests-tree-builder/base.ts
```diff
@@ -184,23 +201,6 @@ export class BaseTestsTreeBuilder {
         }
     }
 
-    protected _addNodeId(parentSuiteId: string, nodeId: string, {fieldName}: {fieldName: 'browserIds' | 'suiteIds'}): void {
```
KuznetsovRoman, line 187: This method is called "addNodeId", so it have to check for duplicates with slow "_isNodeIdExists"
But in "_addSuites" we usually don't need to do it because we process suites from parent to child, and quick "tree.suites.byId[childSuiteId]" is enough

lib/tests-tree-builder/base.ts
```diff
@@ -241,26 +241,59 @@ export class BaseTestsTreeBuilder {
...
+        suite.status = determineFinalStatus(uniqueChildStatuses) || void 0;
 
-        // if newly determined status is the same as current status, do nothing
-        if (suite.status === status) {
+        if (!suite.status) {
             return;
         }
 
-        suite.status = status || undefined;
-        this._setStatusForBranch(testPath.slice(0, -1));
```
KuznetsovRoman, line 264: For each suite we would fully recalculate its parent state
But given that "determineFinalStatus(determineFinalStatus(A, B), C) === determineFinalStatus(A, B, C)", we can only recalculate parent status with "determineFinalStatus(parentStatus, newChildStatus)", because parentStatus is result of "determineFinalStatus" being called on all other nodes

PR #764, opened 2026-04-21
feat: test retry ui

sonic16x:


shadowusr:
@codex review

lib/gui/server.ts
```diff
@@ -160,10 +160,14 @@ export const start = async (args: ServerArgs): Promise<ServerReadyData> => {
...
     });
 
-    server.post('/run', (req, res) => {
+    server.post('/run', async (req, res) => {
         try {
+            const {tests, repeatCount} = req.body;
             // do not wait for completion so that response does not hang and browser does not restart it by timeout
-            app.run(req.body);
+            for (let i = 0; i < repeatCount; i++) {
+                await app.run(tests, {retry: repeatCount === 1});
```
shadowusr: What does passing `retry` change?

Also, the comments in code above and codex's are valid IMO. I've tried running slower tests on a large project and the request takes minutes to complete. I'm afraid it will timeout if I tried to run more tests.

⚠️ Another important issue I've noticed is that the retry button becomes active again after one retry is finished and we are waiting for a session, which is quite misleading, especially when getting a new session takes a while.

shadowusr (approved):
Loooks good to me other than that issue with retry button re-enabling between retries.

lib/gui/server.ts
sonic16x: Fixed both problems

sonic16x: retry need for disable retry for repeat mode

PR #766, opened 2026-05-11
Save report on process exit

sonic16x:
Save report on process exit

image

### DIFF
--- a/testplane.ts
+++ b/testplane.ts
@@ -106,7 +106,7 @@ export default (testplane: Testplane, opts: Partial<ReporterOptions>): void => {
     }));
 
     // wait EXIT event from testplane (it happens when testplane catch process exit)
-    testplane.once(testplane.events.EXIT, async () => {
+    testplane.once(testplane.events.EXIT, () => {
         // then we catch testplane exit(1) call, and now we exactly can save report with all tests
         process.on('exit', async () => {
             try {
--- a/testplane.ts
+++ b/testplane.ts
@@ -64,10 +64,10 @@ export default (testplane: Testplane, opts: Partial<ReporterOptions>): void => {
         });
     });
 
-    const saveReport = async (): Promise<void> => {
+    const saveReport = _.once(async (): Promise<void> => {
         await staticReportBuilder.finalize();
         await htmlReporter.emitAsync(htmlReporter.events.REPORT_SAVED, {reportPath: config.path});
-    };
+    });
 
     testplane.on(testplane.events.INIT, withMiddleware(async () => {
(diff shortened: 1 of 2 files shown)


testplane.ts
```diff
@@ -104,6 +110,12 @@ export default (testplane: Testplane, opts: Partial<ReporterOptions>): void => {
         staticReportBuilder.registerWorkers(createWorkers(runner as unknown as CreateWorkersRunner));
     }));
 
+    testplane.on(testplane.events.EXIT, () => {
```
sonic16x: Need it to be sure that report will be saved after last test failed

testplane.ts
```diff
@@ -104,6 +105,14 @@ export default (testplane: Testplane, opts: Partial<ReporterOptions>): void => {
         staticReportBuilder.registerWorkers(createWorkers(runner as unknown as CreateWorkersRunner));
     }));
 
+    testplane.once(testplane.events.EXIT, async () => {
+        process.on('exit', async () => {
```
KuznetsovRoman: Why are you adding "process exit" event handler on "testplane exit" event?

Maybe you meant to call "saveReport" on "testplane exit"?

Why there are now 2 "saveReport" calls?

sonic16x: Add once and comments

testplane.ts
```diff
@@ -104,6 +105,16 @@ export default (testplane: Testplane, opts: Partial<ReporterOptions>): void => {
         staticReportBuilder.registerWorkers(createWorkers(runner as unknown as CreateWorkersRunner));
     }));
 
+    // wait EXIT event from testplane (it happens when testplane catch process exit)
+    testplane.once(testplane.events.EXIT, async () => {
```
KuznetsovRoman: Uselsess "async". There is no async in it.

testplane.ts
```diff
@@ -104,6 +105,16 @@ export default (testplane: Testplane, opts: Partial<ReporterOptions>): void => {
...
+        // then we catch testplane exit(1) call, and now we exactly can save report with all tests
+        process.on('exit', async () => {
```
KuznetsovRoman: Missleading "async"
"process exit" is sync event:

> Listener functions **must** only perform **synchronous** operations

```js
const process = require('node:process');

process.on('exit', (code) => {
  setTimeout(() => {
    console.log('This will not run');
  }, 0);
});

```

Source: https://nodejs.org/api/process.html#event-exit

So im not sure if this code is guaranteed to work, considering it has 2 async/await

KuznetsovRoman (changes requested):

testplane.ts
sonic16x: Fixed

PR #769, opened 2026-05-24
Sync filters between Suite and Visual Checks pages

sonic16x:


lib/static/modules/middlewares/local-storage.js
```diff
@@ -21,8 +21,7 @@ export default store => next => action => {
         if (action.type === actionNames.VISUAL_CHECKS_SET_DIFF_MODE) {
             localStorageWrapper.setItem(VISUAL_CHECKS_PAGE_DIFF_MODE_KEY, action.payload.diffModeId);
         }
-        localStorageWrapper.setItem('app.suitesPage.viewMode', app.suitesPage.viewMode);
-        localStorageWrapper.setItem('app.visualChecksPage.viewMode', app.visualChecksPage.viewMode);
+        localStorageWrapper.setItem('app.viewMode', app.viewMode);
```
shadowusr: This doesn't work: after page reload, filter resets to "All" regardless of what was chosen previously

shadowusr (commented):
One more thing that I think belongs to this PR is syncing selected tree items between two pages. So when user opens certain test and has some visual check in view, we should attempt to open it in visual checks, similarly, when selecting any visual checks and then going to suites page, we should open the exact suite&visual check that was selected on the visual checks page.

In case there's no visual check associated with suite on visual checks page, we should do nothing, preserving old behavior.

lib/static/modules/middlewares/local-storage.js
sonic16x: Fixed both problem, now view saved in localStorage (this bug was in master also) and now selected items also synced between pages

shadowusr (approved):
Overall, looks great! Two things:

- Could you also please check one more thing, since you’ve been changing the browser filters state logic: does filtering by browser work on the Visual Checks page at all? It doesn’t work on master, and it looks like it probably never did.

- I’m also still a bit unsure about this flow:

1. Select a test without a visual check.
2. Go to the Visual Checks page, where the first item gets auto-selected.
3. Go back to the Suites page, and we end up on the test that contains the first visual check in the list.

IMO, this feels a bit strange. I would expect this flow to end on the same test that was originally selected. When switching to Visual Checks, we could either show “nothing selected” or avoid treating this auto-selection of the first item as a real user selection.

But this is not critical — please take a look and do what seems best to you.

PR #773, opened 2026-06-02
Refresh tests tree

sonic16x:
Screenshot 2026-06-02 at 19 13 58


lib/gui/tool-runner/index.ts
```diff
@@ -154,6 +154,11 @@ export class ToolRunner {
         return this._toolAdapter.readTests(this._testFiles, this._globalOpts);
     }
 
+    async refreshTests(): Promise<void> {
+        this._ensureReportBuilder().saveDb();
```
KuznetsovRoman: Why do we need to save db before initializing?

Why do "refreshTests" method calls reinit?

Maybe it would be better to pull "what is necessary" from "initialize" to extra method?

Perhaps we could skip saving db this way

lib/static/new-ui/features/suites/components/SuitesPage/index.tsx
```diff
@@ -218,6 +218,12 @@ export function SuitesPage(): ReactNode {
         }
     }, [visibleTreeNodeIds, tree, dispatch]);
 
+    useEffect(() => {
+        if (currentBrowser && !currentBrowserEntity && isInitialized && visibleTreeNodeIds.length > 0) {
+            onSelectFirstResult();
+        }
+    }, [currentBrowser, currentBrowserEntity, isInitialized, visibleTreeNodeIds.length]);
```
KuznetsovRoman, line 225: What is it for?

lib/sqlite-client.ts
```diff
@@ -120,6 +120,10 @@ export class SqliteClient {
         return this._db;
     }
 
+    save(): void {
+        fs.writeFileSync(this._dbPath, this._db.export());
+    }
```
KuznetsovRoman: "close" is sync because we can't call async operations on sync "exit" event

But "save" is what you call from async code. Its better to have it async.

KuznetsovRoman (changes requested):

lib/static/new-ui/features/suites/components/SuitesPage/index.tsx
sonic16x: For select first result in case when current selected test after refresh was deleted

lib/sqlite-client.ts
sonic16x: Fixed

lib/gui/tool-runner/index.ts
sonic16x: We need it for using merge mechanism in initialize. For example, If we run some tests and then refresh tests, first we need save current result and then reinit for read new tests and merge it with current results.
In other case we should realize merge logic second time in frontend.

lib/adapters/tool/testplane/test-results-handler.ts
```diff
@@ -75,7 +77,9 @@ export const handleTestResults = (testplane: TestplaneWithHtmlReporter, reportBu
                 const testBranch = reportBuilder.getTestBranch(formattedResult.id);
 
                 return client.emit(clientEventName, testBranch);
-            }).catch(logError);
+            }).catch(() => {
+                logError(new Error('ololo'));
+            });
```
KuznetsovRoman: ?

lib/adapters/tool/testplane/test-results-handler.ts
```diff
@@ -13,6 +13,8 @@ import {TestplaneWithHtmlReporter} from '../../tool/testplane/index';
 import {copyAndUpdate} from '../../test-result/utils';
 
 export const handleTestResults = (testplane: TestplaneWithHtmlReporter, reportBuilder: GuiReportBuilder, client: EventSource): void => {
+    testplane.removeAllListeners();
```
KuznetsovRoman: It removes all listeners, including these ones:
https://nda.ya.ru/t/nu5584BV7ik3kV

You can't do that.

You have that "event listeners" problem because you call "initialize" second time.
"just removing event listeners, while keeping second init" is sweeping the problem under the carpet.

sonic16x: Fixed

lib/adapters/tool/testplane/test-results-handler.ts
sonic16x: Removed

lib/static/new-ui/types/store.ts
```diff
@@ -260,6 +260,7 @@ export interface State {
         isInitialized: boolean;
         availableFeatures: Feature[],
         isSearchLoading?: boolean;
+        isRefreshLoading?: boolean;
```
KuznetsovRoman: Without this PR context, its unclear, what does "is refresh loading" mean
Maybe we should name it more precise?
Like "are tests refreshing", idk

lib/static/new-ui/components/TreeActionsToolbar/index.tsx
```diff
@@ -192,8 +196,22 @@ export function TreeActionsToolbar({onHighlightCurrentTest, className}: TreeActi
     const pluginComponents = getExtensionPointComponents(loadedPluginConfigs, ExtensionPointName.RunTestOptions);
     const hasRunTestOptions = pluginComponents.length > 0;
 
+    const handleRefresh = useCallback(() => {
+        dispatch(thunkRefreshGuiReport());
```
KuznetsovRoman, line 200: nit: As i remember, we dont usually dispatch actions in components manually with "dispatch" function. Instead, we usually do this:
https://github.com/search?q=repo%3Agemini-testing%2Fhtml-reporter%20actions%3A%20bindActionCreators&type=code

lib/static/modules/actions/lifecycle.ts
```diff
@@ -139,6 +140,25 @@ export const thunkInitStaticReport = ({isNewUi}: InitStaticReportData = {}): App
...
 
+export const thunkRefreshGuiReport = (): AppThunk => {
+    return async (dispatch, getState) => {
+        dispatch(setRefreshLoading(true));
+        try {
+            const {db} = getState();
+            const response = await axios.get<GetInitResponse>('/refresh-tests');
+
+            if (!response.data || !db) {
+                throw new Error('Could not refresh tests data.');
```
KuznetsovRoman: > Could not refresh tests data

Why? What should we do in this case?
More actionable error here would be better.
In what case can we not have db?
In what cases response.data will be falsy?
What the user should do in this case?

At least, if response data is missing, we could also provide something about it (like response code), so it would be easier to debug "why user got `Could not refresh tests data` error"

lib/report-builder/gui.ts
```diff
@@ -76,6 +77,12 @@ export class GuiReportBuilder extends StaticReportBuilder {
         };
     }
 
+    resetTree(): void {
+        this._testsTree = GuiTestsTreeBuilder.create({baseHost: this._reporterConfig.baseHost});
+        this._testAttemptManager = new TestAttemptManager();
```
KuznetsovRoman: Test attempt manager lives on `StaticReportBuilder` abstraction layer.

I would suggest not to bring it on `GuiReportBuilder` layer.

We could do it implementing `TestAttemptManager`-s `reset` method, for example

Or at least implementing "reset test attempt manager" method on `StaticReportBuilder` layer

lib/gui/tool-runner/index.ts
```diff
@@ -154,6 +154,16 @@ export class ToolRunner {
         return this._toolAdapter.readTests(this._testFiles, this._globalOpts);
     }
 
+    async refreshTests(): Promise<void> {
+        this._collection = await this._readTests();
+
+        const reportBuilder = this._ensureReportBuilder();
+        reportBuilder.resetTree();
```
KuznetsovRoman: nit: `_ensureReportBuilder` returns report builder instance for convenient chaining pattern:

```js 
this._ensureReportBuilder().resetTree();
```

- https://github.com/gemini-testing/html-reporter/blob/4bfcd5dd04fbe1256849ebf16c0ed7edfc94e7a3/lib/gui/tool-runner/index.ts#L186

- https://github.com/gemini-testing/html-reporter/blob/4bfcd5dd04fbe1256849ebf16c0ed7edfc94e7a3/lib/gui/tool-runner/index.ts#L174

KuznetsovRoman (approved):

PR #777, opened 2026-07-06
fix: show text correctly in "show tests" select in old ui

DudaGod:


lib/static/components/controls/common-controls.jsx
```diff
@@ -52,7 +52,7 @@ class ControlButtons extends Component {
                 <ControlSelect
                     size='s'
                     label="Show tests"
-                    value={app.suitesPage.viewMode}
+                    value={app.viewMode}
```
DudaGod, line 55: the problem appeared after this PR - https://github.com/gemini-testing/html-reporter/pull/769

KuznetsovRoman (approved):

PR #779, opened 2026-07-09
Fix accept visible images

sonic16x:
Screenshot 2026-07-10 at 02 16 46

Now, if image not expanded (`check 4` on screenshot) it will not be accepted

### DIFF
--- a/lib/static/modules/selectors/tree.js
+++ b/lib/static/modules/selectors/tree.js
@@ -192,10 +192,17 @@ export const getSelectedBrowserIds = createSelector(
     }
 );
 
-const getAcceptableOrRevertableImages = (browserIds, browsersById, resultsById, imagesById, isStaticImageAccepterEnabled, gui) => {
+const getAcceptableOrRevertableImages = (
+    browserIds, browsersById, resultsById, imagesById, isStaticImageAccepterEnabled, gui, expandedTreeNodesById
+) => {
     const visibleResultIds = browserIds.map(browserId => last(browsersById[browserId].resultIds));
 
-    return visibleResultIds.flatMap(resultId => {
+    return visibleResultIds.flatMap((resultId) => {
+        const suitePath = resultsById[resultId].suitePath.join('/');
+        if (expandedTreeNodesById && !expandedTreeNodesById[suitePath]) {
+            return [];
+        }
+
         return resultsById[resultId].imageIds
             .map(imageId => imagesById[imageId])
             .filter(image => isAcceptable(image) || isScreenRevertable({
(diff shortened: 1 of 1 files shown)


lib/static/modules/selectors/tree.js
```diff
@@ -192,10 +192,17 @@ export const getSelectedBrowserIds = createSelector(
...
 
-const getAcceptableOrRevertableImages = (browserIds, browsersById, resultsById, imagesById, isStaticImageAccepterEnabled, gui) => {
+const getAcceptableOrRevertableImages = (
+    browserIds, browsersById, resultsById, imagesById, isStaticImageAccepterEnabled, gui, expandedTreeNodesById
+) => {
     const visibleResultIds = browserIds.map(browserId => last(browsersById[browserId].resultIds));
 
-    return visibleResultIds.flatMap(resultId => {
+    return visibleResultIds.flatMap((resultId) => {
+        const suitePath = resultsById[resultId].suitePath.join('/');
```
KuznetsovRoman: This `.join("/")` looks "magic"

`expandedTreeNodesById`, looking at its name, wants `Id` as key

We need `suiteId`

SuiteId can be derived from result like this:
```js
  const result = state.tree.results.byId[resultId];
  const browser = state.tree.browsers.byId[result.parentId];
  const suiteId = browser.parentId;
  // const suite = state.tree.suites.byId[suiteId];
```

lib/static/modules/selectors/tree.js
```diff
@@ -209,6 +216,7 @@ const getAcceptableOrRevertableImages = (browserIds, browsersById, resultsById,
 
 export const getVisibleImages = createSelector(
     getVisibleBrowserIds, getBrowsers, getResults, getImages, getIsStaticImageAccepterEnabled, getIsGui,
+    (state) => state.ui.suitesPage.expandedTreeNodesById,
```
KuznetsovRoman: Sinse other arguments are named functions, i think we should pull this `(state) => state.ui.suitesPage.expandedTreeNodesById` into named function (like getExpandedTreeNodesById)

KuznetsovRoman (changes requested):

lib/static/modules/selectors/tree.js
sonic16x: Screenshot 2026-07-14 at 12 35 56

No, in expandedTreeNodesById we have object like this (see above), it is entityId from useList hook (from gravity-ui)

lib/static/modules/selectors/tree.js
sonic16x: Fixed

lib/static/modules/selectors/tree.js
KuznetsovRoman: A little helper to get "treeNodeId" from result would be nice, so we would call "getTreeNodeId" everywhere we need, without having to remember "its constructed as `suitePath.join('/')`"

KuznetsovRoman (approved):

PR #780, opened 2026-07-10
Feat copy command args

sonic16x:
Screenshot 2026-07-10 at 18 33 48
Screenshot 2026-07-10 at 18 34 16

### DIFF
--- a/lib/static/new-ui/features/suites/components/TestStepArgs/index.module.css
+++ b/lib/static/new-ui/features/suites/components/TestStepArgs/index.module.css
@@ -4,7 +4,19 @@
     overflow: hidden;
 }
 
+.copy {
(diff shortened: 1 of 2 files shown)


shadowusr (approved):
LGTM!

A tiiiny visual enhancement I would do is adding box-shadow, so that the "copy" icon ended not abruptly, but with a soft transition:
Image

But that's 100% non critical thing.

Another thing I want to note is colours that you use in TestStepArgs. Those all are hard-coded values, no problem for now, but when you'll return to the "dark mode" PR, these values would surface. Up to you — whether to make them theme-configurable now or later.

Also, adding e2e on this wouldn't hurt (and usually takes very little time to generate with LLMs), but that's up to you.

PR #783, opened 2026-07-20
Add image res in tree

sonic16x:
Screenshot 2026-07-20 at 12 39 54


lib/static/new-ui/components/TreeViewItemSubtitle/index.tsx
```diff
@@ -55,16 +61,23 @@ export function TreeViewItemSubtitle(props: TreeViewItemSubtitleProps): ReactNod
...
+                        if (imageEntity.differentPixels && imageEntity.diffRatio) {
+                            subTitle = `${getDisplayedDiffPixelsCountValue(imageEntity.differentPixels)} ⋅ ${getDisplayedDiffPercentValue(imageEntity.diffRatio)}%`;
+                        }
                         images.push({
                             title: 'Diff',
+                            subTitle,
                             image: imageEntity.diffImg
                         });
                     }
+                    console.log('DEBUG', props.item);
```
shadowusr: Debug log

shadowusr (approved):
Debug log needs to be removed, and I'd probably make these labels consistent with how we render them on the right side — in a slightly less active color:

Image

But that's up to you. 

Other than that, LGTM!

PR #784, opened 2026-07-20
Disable magnifier in tree

sonic16x:
Screenshot 2026-07-20 at 12 41 47


shadowusr (approved):
Disabling looks fine to me!

However, currently it only disables the magnifier feature, but doesn't bring much value to the user. I think, in the turned off state, the "diff bubbles" feature should become available. Judging by the code, it should be simple enough to implement — just pass the "diff regions" field for the diff image to <Screenshot/> component.

Other than that, all good!

PR #786, opened 2026-07-20
Fix refresh test tree

sonic16x:


lib/gui/tool-runner/index.ts
```diff
@@ -157,10 +157,15 @@ export class ToolRunner {
...
-        this._ensureReportBuilder().resetTree();
+        const reportBuilder = this._ensureReportBuilder();
+        reportBuilder.resetTree();
         this._testAdapters = {};
 
         await this._handleRunnableCollection();
+
+        reportBuilder.reuseTestsTree(reportBuilder.buildTreeFromCurrentDb());
+        const {autoRun} = this._guiOpts;
+        this._tree = {...reportBuilder.getResult(), autoRun, browserFeatures: {}, features: []};
```
KuznetsovRoman: Why do we nullify the features?

lib/report-builder/gui.ts
```diff
@@ -82,6 +84,27 @@ export class GuiReportBuilder extends StaticReportBuilder {
         this.resetAttemps();
     }
 
+    buildTreeFromCurrentDb(): Tree {
+        const db = this._dbClient.getRawConnection();
+        const testsTreeBuilder = StaticTestsTreeBuilder.create({baseHost: this._reporterConfig.baseHost});
+
+        const suitesQueryStatement = db.prepare(commonSqliteUtils.selectAllSuitesQuery());
```
KuznetsovRoman: Can't do that. Facade pattern violation.

If you want some kind of sql logic, either use existing "sqlite-client.ts" method (besides `getRawConnection`), either create a new one

KuznetsovRoman (changes requested):

lib/report-builder/gui.ts
sonic16x: Fixed, created new method in sqlite-client.ts

lib/gui/tool-runner/index.ts
sonic16x: removed, now I reuse _fillTestsTree method (this logic was from here)

lib/report-builder/gui.ts
```diff
@@ -82,6 +82,13 @@ export class GuiReportBuilder extends StaticReportBuilder {
         this.resetAttemps();
     }
 
+    buildTreeFromCurrentDb(): Tree {
+        const testsTreeBuilder = StaticTestsTreeBuilder.create({baseHost: this._reporterConfig.baseHost});
```
KuznetsovRoman, line 86: Lets use `GuiTestsTreeBuilder` here instead of `StaticTestsTreeBuilder`

It should be easy 1:1 replacement and `StaticTestsTreeBuilder` import would be obsolete: as it should be in "report-builder/**_gui_**.ts"

KuznetsovRoman (approved):

lib/report-builder/gui.ts
sonic16x: No, we need build method from StaticTestsTreeBuilder

PR #789, opened 2026-07-21
Fix error view

sonic16x:
Now we save state in local storage. Also I add change icon for button when we switch break lines.

### DIFF
--- a/lib/static/new-ui/components/ErrorInfo/index.tsx
+++ b/lib/static/new-ui/components/ErrorInfo/index.tsx
@@ -1,12 +1,14 @@
 import ansiHtml from 'ansi-html-community';
 import classNames from 'classnames';
 import escapeHtml from 'escape-html';
-import React, {ReactNode, useMemo, useState} from 'react';
+import React, {ReactNode, useMemo} from 'react';
 import {ClipboardButton, Button} from '@gravity-ui/uikit';
 
 import styles from './index.module.css';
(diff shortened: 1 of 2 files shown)


lib/static/new-ui/components/ErrorInfo/index.tsx
```diff
@@ -50,9 +53,13 @@ export function ErrorInfo(props: ErrorInfoProps): ReactNode {
...
                     onClick={(): void => setBreakLines(!breakLines)}
                 >
-                    <Button.Icon>
-                        <img src={WordWrapIcon} width={16} height={16} alt=""/>
-                    </Button.Icon>
+                    {breakLines ? (
+                        <Icon data={Bars} width={16} height={16}/>
+                    ) : (
+                        <Button.Icon>
+                            <img src={WordWrapIcon} width={18} height={18} alt=""/>
```
KuznetsovRoman: Why sizes did change? 16x16 -> 18x18
`Icon` above is 16x16

KuznetsovRoman (approved):

lib/static/new-ui/components/ErrorInfo/index.tsx
sonic16x: Removed logic witch change icon (in doc we change color not icon, but here it looks strange) but keep 18px size, because it is the same size which have copy icon

PR #793, opened 2026-08-02
Remove unnecessary spinners

sonic16x:
Screenshot 2026-08-03 at 00 28 50


shadowusr (approved):
I would remove this spinner as well:

Image

Tiny issue with Retry button — it still changes width slightly when switching between "Retry"  and "Running" states. idk, maybe just disable it without changing text. Not critical.

PR #795, opened 2026-08-06
Add run options str

sonic16x:
Screenshot 2026-08-06 at 15 25 33
Screenshot 2026-08-06 at 15 25 43
Screenshot 2026-08-06 at 15 25 55


lib/static/modules/reducers/running.js
```diff
@@ -9,6 +9,10 @@ export default (state, action) => {
         case actionNames.SET_REPEAT_LEFT: {
             return {...state, repeatLeft: action.payload.repeatLeft};
         }
+        case actionNames.SET_RUN_OPTIONS: {
+            console.log('DEBUG payload', action.payload.runOptions);
```
shadowusr, line 13: debug logging

lib/static/modules/actions/run-tests.ts
```diff
@@ -8,11 +8,13 @@ import {connectToDatabase, getMainDatabaseUrl} from '@/db-utils/client';
 import {createNotificationError} from '@/static/modules/actions/notifications';
 import {TestBranch} from '@/tests-tree-builder/gui';
 import {TestStatus} from '@/constants';
+import {RunOptions} from '@/static/new-ui/types/store';
 
 export type RunTestAction = Action<typeof actionNames.RETRY_TEST>;
 export const runTest = (): RunTestAction => ({type: actionNames.RETRY_TEST});
 export const setRepeatCount = (repeatCount: number): Action<typeof actionNames.SET_REPEAT_COUNT, {repeatCount: number}> => ({type: actionNames.SET_REPEAT_COUNT, payload: {repeatCount}});
 export const setRepeatLeft = (repeatLeft: number): Action<typeof actionNames.SET_REPEAT_LEFT, {repeatLeft: number}> => ({type: actionNames.SET_REPEAT_LEFT, payload: {repeatLeft}});
+export const setRunOptions = (runOptions: RunOptions): Action<typeof actionNames.SET_RUN_OPTIONS, {runOptions: RunOptions}> => ({type: actionNames.SET_RUN_OPTIONS, payload: {runOptions}});
```
shadowusr, line 17: Hmmm, so with this action, every time someone wants to update options, it's first needed to get them, merge and update. What if we had something like this:

```typescript
type RunOptions = Record<string, string[]>;

const setRunOptions = (namespace: string, options: string[]) => /* ... */
```

This would allow consumers to conveniently update only parts of the state they care about. And having an array instead of `string | null` allows to both avoid dealing with nulls and have multiple values just in case.

What do you think?

lib/static/new-ui/hooks/useRunOptionsString.ts
```diff
@@ -0,0 +1,17 @@
+import {useSelector} from 'react-redux';
+
+export const useRunOptionsString = (): string => useSelector((state) => {
+    const opts: string[] = [];
+
+    if (state.repeatCount > 1) {
+        opts.push(`x${state.repeatCount}`);
```
shadowusr, line 7: Let's use the \times symbol here as well: `×`

lib/static/new-ui/hooks/useRunOptionsString.ts
```diff
@@ -0,0 +1,17 @@
...
+    }
+
+    Object.values(state.runOptions).forEach((value) => {
+        if (value && value.length > 0) {
+            opts.push(value);
+        }
+    });
+
+    return opts.join('⋅');
```
shadowusr, line 16: nit: maybe let's add spaces around the dot? but that's up to you

shadowusr (approved):

shadowusr:
Also, forgot to mention, let's make the text on the run button slightly brighter. I think we can change opacity from `0.5` to `0.7`.

PR #800, opened 2026-08-23
Feat stop tests button

sonic16x:
Now user can stop all test after run using stop button which appear after click to start button.

Screenshot 2026-08-26 at 04 20 26

Screenshot 2026-08-23 at 23 22 23

Screenshot 2026-08-26 at 04 20 50

### DIFF
--- a/lib/static/new-ui/components/MainLayout/hotkeys.ts
+++ b/lib/static/new-ui/components/MainLayout/hotkeys.ts
@@ -28,6 +28,7 @@ export const HOTKEYS_GROUPS: HotkeysGroup[] = [
             {title: 'Next attempt', value: '→'},
             {title: 'Run current test', value: 'r'},
             {title: 'Run all/selected tests', value: 'shift+r'},
+            {title: 'Stop all tests', value: 's'},
             {title: 'Accept screenshot', value: 'a'},
             {title: 'Undo accept', value: 'u'},
             {title: 'Accept all/selected', value: 'shift+a'},
--- a/lib/static/new-ui/components/MainLayout/hotkeys.ts
+++ b/lib/static/new-ui/components/MainLayout/hotkeys.ts
@@ -28,7 +28,7 @@ export const HOTKEYS_GROUPS: HotkeysGroup[] = [
             {title: 'Next attempt', value: '→'},
             {title: 'Run current test', value: 'r'},
             {title: 'Run all/selected tests', value: 'shift+r'},
-            {title: 'Stop all tests', value: 's'},
+            {title: 'Stop all tests', value: 'shift+s'},
             {title: 'Accept screenshot', value: 'a'},
             {title: 'Undo accept', value: 'u'},
             {title: 'Accept all/selected', value: 'shift+a'},
--- a/lib/adapters/test-result/transformers/db.ts
+++ b/lib/adapters/test-result/transformers/db.ts
@@ -16,6 +16,16 @@ export class DbTestResultTransformer {
 
     transform(testResult: ReporterTestResult): DbTestResult {
         const suiteUrl = getUrlWithBase(testResult.url, this._options.baseHost);
+        const imagesInfo = (testResult.imagesInfo ?? []).map(imageInfo => {
+            if (!_.isObject(imageInfo) || !('error' in imageInfo) || !imageInfo.error) {
+                return imageInfo;
+            }
+
+            return {
+                ...imageInfo,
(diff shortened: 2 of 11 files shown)


lib/static/new-ui/components/MainLayout/hotkeys.ts
```diff
@@ -28,6 +28,7 @@ export const HOTKEYS_GROUPS: HotkeysGroup[] = [
             {title: 'Next attempt', value: '→'},
             {title: 'Run current test', value: 'r'},
             {title: 'Run all/selected tests', value: 'shift+r'},
+            {title: 'Stop all tests', value: 's'},
```
shadowusr: We have a conflict, like 20 lines above — `S` is already used for Suites page.

shadowusr:
Browsers are not closed reliably on stop, I reproduced this multiple times, when after stopping, the browsers are left hanging:

Screenshot 2026-08-24 at 5 02 53 PM

shadowusr:
It doesn't seem to work right with multiple retries configured in UI. I tested it together with these changes: https://github.com/gemini-testing/testplane/pull/1318

And yet, after clicking the stop button, it will stop only the current retry and not subsequent ones.

shadowusr:
IMO this looks pretty ugly:

Screenshot 2026-08-24 at 5 40 59 PM

It would be really nice if we didn't show any error at all in this case, but instead clearly stated in status "Interrupted" and for example had gray icon and gray retry selector background (not red).

This is not absolutely critical, but would be very nice to have.

shadowusr:
As for the button on the right: I think we should be able to stop tests from there. We should just clearly state "Stop all" if stopping individual test is currently hard to implement. I would imagine that upon clicking on "Retry" button it would be replaced with "Stop All" button, that doesn't have arrow down with options on the right and has the exact same size to avoid UI jumps

shadowusr:
I think e2e test on this is a must. Especially considering that it would be very easy to set it up: we already have GUI tests, so any LLM would easily generate this one.

lib/static/new-ui/components/MainLayout/hotkeys.ts
sonic16x: Now `shitf+s`, but maybe just `x` ?

sonic16x:
Screenshot 2026-08-26 at 04 57 50

Maybe red background? But I'm not sure

shadowusr:
It works much better now, but still occasionally leaves opened browsers behind (but that's rare and not critical). When stopping tests that are waiting for session from a remote grid, i get these errors:
```
CancelledError
CancelledError: Browser request was cancelled

What happened:
- This test tried to run in a browser that was already stopped
- This likely happened due to a critical error, like an unhandled promise rejection
What you can do:
- Check other failed tests or execution logs for more details, usually you can find the root cause there
    at BasicPool.getBrowser (/Users/shadowusr/dev/testplane/wt/fix-testplane-9-on-old-browsers/build/src/browser-pool/basic-pool.js:79:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async LimitedPool.getBrowser (/Users/shadowusr/dev/testplane/wt/fix-testplane-9-on-old-browsers/build/src/browser-pool/limited-pool.js:37:20)
    at async BrowserAgent.getBrowser (/Users/shadowusr/dev/testplane/wt/fix-testplane-9-on-old-browsers/build/src/runner/browser-agent.js:13:25)
    at async RegularTestRunner._getBrowser (/Users/shadowusr/dev/testplane/wt/fix-testplane-9-on-old-browsers/build/src/runner/test-runner/regular-test-runner.js:155:29)
    at async RegularTestRunner._run (/Users/shadowusr/dev/testplane/wt/fix-testplane-9-on-old-browsers/build/src/runner/test-runner/regular-test-runner.js:47:29)
    at async InsistantTestRunner.run (/Users/shadowusr/dev/testplane/wt/fix-testplane-9-on-old-browsers/build/src/runner/test-runner/insistant-test-runner.js:38:9)
    at async BrowserRunner._runTest (/Users/shadowusr/dev/testplane/wt/fix-testplane-9-on-old-browsers/build/src/runner/browser-runner.js:102:9)
```

Instead of clean interrupted by user error. I think this would be a frequent case for sessions that take a while to start, for example, android sessions. So it would be nice to be fixed.

Also, I think it would be very important to check how it works with remote grid in this case: will those sessions be left hanging or will they be correctly stopped?

shadowusr (approved):

PR #804, opened 2026-09-02
Feat watch mode

sonic16x:
Watch mode for gui mode.

Usage:
`npx testplane gui --watch`

https://github.com/user-attachments/assets/221eef0f-0f4b-4b54-a349-f9ef2b4b3373


lib/gui/server.ts
```diff
@@ -295,6 +311,118 @@ export const start = async (args: ServerArgs): Promise<ServerReadyData> => {
 
     await app.initialize();
 
+    if (args.cli.options.watch && toolAdapter.toolName === ToolName.Testplane) {
```
KuznetsovRoman, line 302: Its not the clean architecture i would like to see in html-reporter.
We have a lot of layers in html-reporter (server launches app, which manages tool-runner) to distribute responsibility between the layers, and you just pasted entire watcher in "gui/server.ts".

Now "server.ts" handles chokidar, calculates scope, manages debounce, drain the que, manages this "relatively low" level error handling...

Its clearly should be another module, and not "server.ts" responsibility.

lib/gui/server.ts
```diff
@@ -295,6 +311,118 @@ export const start = async (args: ServerArgs): Promise<ServerReadyData> => {
...
+        const config = toolAdapter.config as TestplaneConfigAdapter;
+        const watchPaths = [...new Set([...config.getTestFilePatterns(), ...args.paths])];
+        const watchRoots = [...new Set(watchPaths.map(getWatchRoot).filter(Boolean))];
```
KuznetsovRoman: You are building watch roots, that may differ from real testplane scope:
- If we have empty default set with `files: []`, your code gets empty array, while testplane reads default test directories
- When running `npx testplane gui tests/a.ts --watch`, watches subscribes on all of the config globs, which is not what should be done here.
- Selecting set through CLI call also is not considered
- If path is set like "tests" without ending "/" or asterisks on the end, your "path.dirname('tests')" returns ".", so the watcher looks at all project files, including node_modules

Examples:
• `getWatchRoot('tests')` -> `.`
• `getWatchRoot('tests/')` -> `tests`
• `getWatchRoot('tests/**/*.ts')` -> `tests`

As a solution i see "creating watch plan in testplane adapter the same way test set is built for initial read". So it would be single source of truth, and not "couple of testfile finding mechanisms, which can work differently"

lib/gui/server.ts
```diff
@@ -295,6 +311,118 @@ export const start = async (args: ServerArgs): Promise<ServerReadyData> => {
...
+        let refreshInProgress = false;
+        const queuedFiles = new Set<string>();
+        const queuedRemovedDirectories = new Set<string>();
+        const debounceFiles = new Set<string>();
+        const debounceRemovedDirectories = new Set<string>();
+        let refreshTimer: NodeJS.Timeout | undefined;
+        let refreshSequence = 0;
+        let firstDebouncedEventAt: number | undefined;
+
+        const refresh = async (changedFiles: string[], removedDirectories: string[]): Promise<void> => {
```
KuznetsovRoman: There is a "catch" (thats a wordplay) in this function:
If some step, like, retrieving sqlite from history, fails, browser would only get "TESTS_REFRESH_FAILED" without actual tree patch. So if there was test A, which was renamed to test B, after the error browser would see test A, not test B, despite server collection would have the opposite: test B, not test A.

Trying to run test A would likely result in silent "nothing is really launched"

lib/gui/tool-runner/index.ts
```diff
@@ -155,16 +176,250 @@ export class ToolRunner {
...
+            return relativePath !== '' && !relativePath.startsWith(`..${path.sep}`) && relativePath !== '..' && !path.isAbsolute(relativePath);
+        });
+        const affectedFiles = new Set(normalizedFiles);
+        for (const testFile of this._testsByFile.keys()) {
+            if (isInsideRemovedDirectory(testFile)) {
+                affectedFiles.add(testFile);
+            }
+        }
+        const isChangedFile = (test: TestAdapter): boolean => affectedFiles.has(path.resolve(test.file));
+        const signature = (test: TestAdapter): string => JSON.stringify([test.browserId, path.resolve(test.file), test.titlePath]);
```
KuznetsovRoman: Your signature only includes "browserId", "resolved file" and "titlePath"
But tree building also consider "disabled", "silentlySkipped" and "pending"
For example, changing "it" to "it.skip" would have the exact signature and therefore would not refresh.

lib/gui/tool-runner/index.ts
```diff
@@ -155,16 +176,250 @@ export class ToolRunner {
...
+        const existingFiles = await Promise.all(changedFiles.map(async file => await fs.pathExists(file) ? file : null));
+        const filesToRead = existingFiles.filter((file): file is string => Boolean(file));
+        logWatchPerformance(performanceId, 'check changed files existence', stageStartedAt, {filesToRead: filesToRead.length});
+        let next: string[] = [];
+        let changedCollection: TestCollectionAdapter = {tests: []};
+
+        if (filesToRead.length) {
+            try {
+                stageStartedAt = performance.now();
+                changedCollection = await this._toolAdapter.readTests(filesToRead, this._globalOpts);
```
KuznetsovRoman, line 251: Shouldn't you separately check for "unique full name" here?
"readTests" under the hood checks for unique full names only for tests it currently read.
But you read only subset of tests and then combine it with previous tree without checking for fullName duplicates. It could potentially lead into some hard-to-debug problem.

lib/static/modules/search/worker.ts
```diff
@@ -86,10 +87,49 @@ type SearchMessage = {
...
     switch (event.data.type) {
         case 'init': {
+            const startedAt = performance.now();
             initSearch(event.data.data);
+            if (event.data.performanceId !== undefined) {
+                console.info(`[watch-perf][client][#${event.data.performanceId}][search worker] rebuild index: ${(performance.now() - startedAt).toFixed(1)}ms`, {
+                    items: Object.keys(event.data.data).length
+                });
+            }
+            self.postMessage(true);
```
KuznetsovRoman: So now worker sends "true" on both "init" and "patch"

Existing "search" function in this file expects array of strings

If we get "true" from worker after search was initiated, "search" could confuse its response as "response to search" and fail due to "TypeError: array expected" or something like that.

I think we should move to typed messages: dispatching not just single "true", but whole object with "type" field in it in order to avoid confusion.

lib/static/modules/search/index.ts
```diff
@@ -1,37 +1,75 @@
...
+        const browser = tree.browsers.byId[browserId];
+        const lastResultId = browser?.resultIds.at(-1);
+        const result = lastResultId ? tree.results.byId[lastResultId] : undefined;
+
+        if (result) {
+            idTagMap[browserId] = getResultTags(result);
+        }
+    });
+
+    worker.postMessage({
```
KuznetsovRoman: Have you checked that "Test DOES appear when there was a filter, preventing from showing it, because its old name did not suite the filter, but the new one - does"?

lib/gui/server.ts
```diff
@@ -295,6 +311,118 @@ export const start = async (args: ServerArgs): Promise<ServerReadyData> => {
...
+                }
+            } catch (error) {
+                app.sendClientEvent(ClientEvents.TESTS_REFRESH_FAILED, undefined);
+                logger.error(`Error while refreshing tests after file change: ${(error as Error).message}`);
+            } finally {
+                refreshInProgress = false;
+            }
+        };
+
+        testsWatcher = chokidar.watch(watchPaths, {
```
KuznetsovRoman: So, "chokidar.watch" errors crash the whole gui process. Is it expected? Is it like super stable so we wont expect it ever happening?

I would add some kind of "catch", at least for easier debugging (when somebody would see scary error message from chokidar.watch exception, he should at least understand that the error was because of some king of error in tree watcher)

lib/sqlite-client.ts
```diff
@@ -172,6 +178,28 @@ export class SqliteClient {
...
     }
 
+    getSuitesByTests(tests: TestHistorySpec[]): RawSuitesRow[] {
+        const rows: RawSuitesRow[] = [];
+        const uniqueTests = _.uniqBy(tests, ({suitePath, browserId}) => `${JSON.stringify(suitePath)}\0${browserId}`);
+
+        for (const {suitePath, browserId} of uniqueTests) {
+            const statement = this._db.prepare(
+                `SELECT * FROM ${DB_SUITES_TABLE_NAME} WHERE suitePath = ? AND name = ?`
+            );
```
KuznetsovRoman: Why are you doing N full scans? Its not clickhouse with its magic data skipping indexes. Its better to iterate through every "suite" once and filter everything you need by yourself, instead of torturing SQLite like that.

KuznetsovRoman (changes requested):
I would appreciate more modular structure, where tests watcher would be pulled to separate module with short and concise module calls instead of being splattered all over the code base.

lib/gui/server.ts
sonic16x: Fixed. Moved watcher logic to separate TestsWatcher module. Now server.ts only creates and starts it.

lib/gui/server.ts
sonic16x: Fixed. Watch plan is now created in Testplane adapter. It handles CLI paths, selected sets, environment sets and default folders. Added tests for this cases.

lib/gui/server.ts
sonic16x: Fixed. Collection is updated only after successful tree update. If update fails, previous tree, test adapters and attempts are restored. Added test for this case.

lib/gui/tool-runner/index.ts
sonic16x: Fixed

lib/gui/tool-runner/index.ts
sonic16x: Fixed. Partial update now checks duplicate test names in changed and unchanged files. Added test for this case.

lib/static/modules/search/index.ts
sonic16x: Fixed. Worker messages now have type and requestId, so different responses are not mixed. Added test for this case.

lib/gui/server.ts
sonic16x: Fixed. Current search filter runs again after search index update. Added test for renamed test case.

lib/sqlite-client.ts
sonic16x: Fixed

lib/gui/tests-watcher/index.ts
```diff
@@ -0,0 +1,174 @@
...
+        this._reportPath = reportPath;
+    }
+
+    start(): void {
+        try {
+            this._testsWatcher = chokidar.watch(this._plan.paths, {
+                cwd: process.cwd(),
+                ignoreInitial: true,
+                ignored: [
+                    /(^|[/\\])\../,
```
KuznetsovRoman, line 59: Looks like it would ignore `.tests/case.ts`

We actually have some concrete examples: https://nda.ya.ru/t/VVkNi_VO7qaK7o

lib/gui/tool-runner/index.ts
```diff
@@ -61,6 +69,25 @@ export interface RunParams {
...
+    error instanceof Error && error.message.startsWith('There are no tests found');
+
+const getTestStructureSignature = (test: TestAdapter): string => JSON.stringify([
+    test.browserId,
+    path.resolve(test.file),
+    test.titlePath,
+    test.disabled,
+    test.silentlySkipped,
+    test.pending
+]);
```
KuznetsovRoman, line 89: Maybe we should also add skip reason here?
Now skip reason update would not update real test in the tree because its not a part of the key

lib/gui/tests-watcher/index.ts
```diff
@@ -0,0 +1,174 @@
...
+                    treeUpdate.performance = {
+                        id: refreshId,
+                        serverStartedAt,
+                        serverCompletedAt: Date.now()
+                    };
+                    const sendStartedAt = performance.now();
+                    this._app.sendClientEvent(ClientEvents.TESTS_REFRESHED, treeUpdate);
+                    logger.log(`[watch-perf][server][#${refreshId}] serialize/write SSE: ${(performance.now() - sendStartedAt).toFixed(1)}ms`);
+                }
+                logger.log(`[watch-perf][server][#${refreshId}] refresh loop total: ${(performance.now() - refreshStartedAt).toFixed(1)}ms ${JSON.stringify({changed})}`);
```
KuznetsovRoman, line 157: Those type of logs would be removed before merging, right?
If you want to leave them in the code, you should at least hide them in "debug"

KuznetsovRoman (approved):

lib/gui/tests-watcher/index.ts
sonic16x: Yea, I will remove it, it is just for review and tests

PR #806, opened 2026-09-07
Feat show testplane config in ui

sonic16x:
Now you can see testplane config in Info section of sidebar menu.

Screenshot 2026-09-11 at 06 34 00


lib/adapters/tool/testplane/index.ts
```diff
@@ -79,6 +79,7 @@ export class TestplaneToolAdapter implements ToolAdapter {
         this._config = TestplaneConfigAdapter.create(this._tool.config);
         this._browserConfigs = _.map(this._config.browserIds, (id) => this._config.getBrowserConfig(id));
         this._htmlReporter = HtmlReporter.create(this._reporterConfig, {toolName: ToolName.Testplane});
+        this._htmlReporter.testplaneConfig = this._config.getUserConfig();
```
shadowusr: Why do we name this field `testplaneConfig`? I see that currently we support only Testplane, but nothing stops us of adding this feature to other tools as well. I would give it a generic name, like `toolConfig` or `userConfig`, etc.

lib/static/new-ui/components/InfoPanel/ConfigSection/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+        <Dialog
+            open={isOpen}
+            size={'l'}
+            className={styles.dialog}
+            modalClassName="config-modal"
+            contentOverflow={'auto'}
+            hasCloseButton={true}
+            onClose={(): void => setIsOpen(false)}
+        >
+            <Dialog.Header caption={'Testplane config'}/>
```
shadowusr, line 157: This results in an inconsistent text style, let's use either panel heading style or section heading style:

Image

lib/adapters/config/testplane.ts
```diff
@@ -1,7 +1,23 @@
+import path from 'node:path';
 import type {Config} from 'testplane';
 import type {ConfigAdapter} from './';
 import type {TestplaneTestAdapter} from '../test/testplane';
 
+export const maskTokenValues = (value: unknown): unknown => {
```
shadowusr: Hmm, I see a pretty important issue here:
- we only look at the field name over here, "token", specifically
- if user has sensitive data in any other field, they have absolutely zero way to fix that

So, with this update, we unconditionally expose sensitive data in all fields not named "token" and user has no way to fix this.

I see two ways of improving this:
1. Look not only at the field name, but at the contents, too, for example strings containing `y1__`, `AQAD`, etc.

2. Provide a way for users to customize this behavior, either turn it off completely or provide their own sanitizer. We should think hard, what's best in this case: trying to avoid adding too many difficult to use options yet providing an escape hatch if user needs it.

We can also look how other providers solve this, for example, pwt.

And this 100% should be tested on large projects, to check that at least for them, it works out of the box.

lib/adapters/config/testplane.ts
```diff
@@ -29,6 +45,27 @@ export class TestplaneConfigAdapter implements ConfigAdapter {
         return this._config.forBrowser(browserId);
     }
 
+    getUserConfig(): Record<string, unknown> {
```
shadowusr: I think `configPath` should not be displayed as part of the config object. Because it's not actually in user config, it's a side-effect of how we work with config. But it's actually a good idea to display it as a separate field in the dialog!

lib/adapters/config/testplane.ts
```diff
@@ -29,6 +45,27 @@ export class TestplaneConfigAdapter implements ConfigAdapter {
...
+        if (!this._config.configPath) {
+            return {};
+        }
+
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        const configModule: unknown = require(path.resolve(this._config.configPath));
```
shadowusr: Hmmm, this discards all env and CLI args overrides.

In fact, simply `return maskTokenValues(this._config.serialize()) as Record<string, unknown>;` gives a more correct config, doesn't it?

It gives a fully resolved config, with all values filled, including unchanged defaults. Yes, it's pretty long, but probably much more useful? And we could post-process it to make it smaller, for example, by omitting values inside browser specifications if they match root level values, etc.

We can discuss in a team chat, if you want.

lib/static/new-ui/components/InfoPanel/ConfigSection/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+                            title={areAllNestedExpanded ? 'Collapse all' : 'Expand all'}
+                            onClick={(): void => setAllNestedExpanded(!areAllNestedExpanded)}
+                        >
+                            <Button.Icon>
+                                <Icon data={areAllNestedExpanded ? ChevronsCollapseVertical : ChevronsExpandVertical}/>
+                            </Button.Icon>
+                        </Button>
+                    </div>
+                    <div className={classNames(styles.config, {[styles.breakLines]: breakLines})}>
+                        <JsonNode value={config} depth={0} areAllNestedExpanded={areAllNestedExpanded}/>
```
shadowusr, line 184: While I don't suggest fixing it right away, but it's just sad that we have 3 different code block styles across the report: this one, error stack traces, and critical error stack traces :(

At some point, it would be nice to unify them.

shadowusr (changes requested):
Exceptionally wonderful and long-awaited feature! 🔥 

I've listed a couple of issues in comments, though, that would be nice to fix.

shadowusr:
Also, let's add at least a couple of e2e tests on this new feature. They should be honest, representing how users would interact with the app. For example open config and check that it contains expected values.

lib/adapters/tool/testplane/index.ts
sonic16x: Fixed, toolConfig now

lib/static/new-ui/components/InfoPanel/ConfigSection/index.tsx
sonic16x: Fixed

lib/adapters/config/testplane.ts
sonic16x: Fixed, added new func secretConfigFilter, user can configure it, but by default used internal filter for keys token, secret etc...

lib/adapters/config/testplane.ts
sonic16x: First I did it, but config with default values too big, and as I think we want see only user config

lib/adapters/config/testplane.ts
sonic16x: Fixed, now separate filed in ui

test/func/tests/utils.js
```diff
@@ -96,6 +96,16 @@ const waitForFsChanges = async (dirPath, condition = (output) => output.length >
 const interceptClipboard = async (browser) => {
     await browser.execute(() => {
         window.__copiedText = null;
+
+        Object.defineProperty(navigator, 'clipboard', {
```
shadowusr, line 100: hmmm why was this needed?

test/func/tests/common/new-ui/config.testplane.js
```diff
@@ -0,0 +1,42 @@
...
+
+                    const toolConfig = await browser.$('[data-qa="tool-config"]');
+                    await toolConfig.moveTo();
+
+                    const copyConfigButton = await browser.$('[data-qa="copy-config"]');
+                    await copyConfigButton.waitForClickable();
+                    await copyConfigButton.click();
+
+                    const copiedConfig = await getClipboardValue(browser);
+                    expect(copiedConfig.length).toBeGreaterThan(20);
```
shadowusr: Let's include actual value checks here, like "at least has gridUrl with correct value". Almost any string would pass "length greater than 20". We want to assess correctness from a user's PoV and users care about actual contents.

lib/static/new-ui/components/InfoPanel/ConfigSection/index.tsx
shadowusr: Let's make font-weight of the "Config path" subheading 400, to be consistent with settings items.

Also, I'd change font-size of the path to 13px:

Image

lib/constants/defaults.ts
```diff
@@ -8,6 +8,7 @@ export const CIRCLE_RADIUS = 150;
 export const configDefaults: StoreReporterConfig = {
     baseHost: '',
     commandsWithShortHistory: [],
+    secretConfigFilter: null,
```
shadowusr, line 11: Let's not forget to document it over here — https://testplane.io/docs/html-reporter/html-reporter-setup/

shadowusr (approved):

PR #808, opened 2026-09-07
fix(playwright): recognize modern screenshot comparison errors

dimazollo:
Playwright 1.59.1 reports screenshot mismatches as:

```text
Error: expect(page).toHaveScreenshot(expected) failed
```

The adapter only recognizes `Screenshot comparison failed`, so it marks these as general errors even when expected/actual/diff images are attached. Accept can then leave a stale error instead of resolving the visual failure.

This adds support for the newer page/locator `toHaveScreenshot` header. Playwright uses the same header when it cannot take a screenshot, so the adapter also checks for matching expected/actual/diff attachments. The old `Screenshot comparison failed` handling stays unchanged.

Only the adapter and its unit tests change. `toMatchSnapshot`, changes to legacy mixed-error handling, and platform handling are outside this PR.

### Reproduction

With Playwright 1.59.1 and `html-reporter/playwright`:

```ts
import {test, expect} from '@playwright/test';

test('screenshot mismatch', async ({page}) => {
    const color = process.env.CHANGED ? 'blue' : 'red';
    await page.setContent(`<div style="width:120px;height:80px;background:${color}"></div>`);
    await expect(page).toHaveScreenshot('example.png');
});
```

```sh
npx playwright test --update-snapshots
CHANGED=1 npx playwright test
```

Before: `error` / `Error`. After: `fail` / `ImageDiffError`.

### Tests

- All 30 adapter tests pass, including page/locator diffs, capture failures, snapshot attachment matching, unnamed diffs, ANSI formatting, and multiple soft screenshot assertions.
- `npm test` passes on Node 20.19.6: lint, Node and jsdom unit tests, and type checks.

The full GUI e2e suite and interactive Accept/Undo weren't rerun for this isolated fix.

I hereby agree to the terms of the CLA available at: https://yandex.ru/legal/cla/?lang=en.

### DIFF
--- a/lib/adapters/test-result/playwright.ts
+++ b/lib/adapters/test-result/playwright.ts
@@ -212,7 +212,8 @@ export class PlaywrightTestResultAdapter implements ReporterTestResult {
 
             if (/snapshot .*doesn't exist/.test(message) && message.includes('.png')) {
                 result.name = ErrorName.NO_REF_IMAGE;
-            } else if (this._testResult.errors.every(error => this._isScreenshotComparisonError(error))) {
+            } else if (message.includes('Screenshot comparison failed') ||
+                this._testResult.errors.every(error => this._isScreenshotComparisonError(error))) {
                 result.name = ErrorName.IMAGE_DIFF;
             }
 
@@ -371,11 +372,7 @@ export class PlaywrightTestResultAdapter implements ReporterTestResult {
     private _isScreenshotComparisonError(error: PlaywrightTestResult['errors'][number]): boolean {
         const message = stripAnsi(error.message || '');
         const header = message.split('\n')[0];
-        if (header.includes('Screenshot comparison failed')) {
-            return true;
-        }
-
-        if (!/^(?:Error: )?expect\((?:page|locator|Buffer)\)\.(?:toHaveScreenshot|toMatchSnapshot)\(expected\)(?: failed)?$/.test(header)) {
+        if (!/^(?:Error: )?expect\((?:page|locator)\)\.toHaveScreenshot\(expected\) failed$/.test(header)) {
             return false;
         }
 
--- a/lib/adapters/test-result/playwright.ts
+++ b/lib/adapters/test-result/playwright.ts
@@ -212,7 +212,7 @@ export class PlaywrightTestResultAdapter implements ReporterTestResult {
 
             if (/snapshot .*doesn't exist/.test(message) && message.includes('.png')) {
                 result.name = ErrorName.NO_REF_IMAGE;
-            } else if (message.includes('Screenshot comparison failed')) {
+            } else if (this._testResult.errors.every(error => this._isScreenshotComparisonError(error))) {
                 result.name = ErrorName.IMAGE_DIFF;
             }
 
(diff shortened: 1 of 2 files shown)


lib/adapters/test-result/playwright.ts
```diff
@@ -212,7 +212,7 @@ export class PlaywrightTestResultAdapter implements ReporterTestResult {
 
             if (/snapshot .*doesn't exist/.test(message) && message.includes('.png')) {
                 result.name = ErrorName.NO_REF_IMAGE;
-            } else if (message.includes('Screenshot comparison failed')) {
+            } else if (this._testResult.errors.every(error => this._isScreenshotComparisonError(error))) {
```
sipayRT: why is it not enough to just switch to 2 error options? for example, `/Screenshot comparison failed|expect\(.*\)\.toHaveScreenshot\(expected\)failed/`

lib/adapters/test-result/playwright.ts
```diff
@@ -368,6 +368,27 @@ export class PlaywrightTestResultAdapter implements ReporterTestResult {
...
     }
 
+    private _isScreenshotComparisonError(error: PlaywrightTestResult['errors'][number]): boolean {
+        const message = stripAnsi(error.message || '');
+        const header = message.split('\n')[0];
+        if (header.includes('Screenshot comparison failed')) {
+            return true;
+        }
+
+        if (!/^(?:Error: )?expect\((?:page|locator|Buffer)\)\.(?:toHaveScreenshot|toMatchSnapshot)\(expected\)(?: failed)?$/.test(header)) {
```
sipayRT: `toMatchSnapshot` should be used for [non-image comparison](https://playwright.dev/docs/test-snapshots#non-image-snapshots), so we don't need this part here

sipayRT (changes requested):

lib/adapters/test-result/playwright.ts
dimazollo: It can compare images too, but I agree that we don't need it for this fix. I'll remove it from this PR.

lib/adapters/test-result/playwright.ts
dimazollo: Playwright uses the same header when it cannot take a screenshot. So I also check that expected, actual, and diff images are attached.

I'll keep `toHaveScreenshot` support with this check and remove the mixed-error changes from this PR.

lib/adapters/test-result/playwright.ts
```diff
@@ -368,6 +369,23 @@ export class PlaywrightTestResultAdapter implements ReporterTestResult {
...
+        const message = stripAnsi(error.message || '');
+        const header = message.split('\n')[0];
+        if (!/^(?:Error: )?expect\((?:page|locator)\)\.toHaveScreenshot\(expected\) failed$/.test(header)) {
+            return false;
+        }
+
+        // Modern Playwright uses the same matcher header for diffs and capture errors.
+        const snapshotName = message.match(/^\s*Snapshot: (.+)\.png\s*$/m)?.[1];
+        const states = Object.entries(this._attachmentsByState).filter(([state]) =>
+            snapshotName ? state === snapshotName : this._testResult.errors.length === 1);
```
sipayRT, line 382: with `expect.soft` error length can be more then 1

lib/adapters/test-result/playwright.ts
```diff
@@ -212,7 +212,8 @@ export class PlaywrightTestResultAdapter implements ReporterTestResult {
 
             if (/snapshot .*doesn't exist/.test(message) && message.includes('.png')) {
                 result.name = ErrorName.NO_REF_IMAGE;
-            } else if (message.includes('Screenshot comparison failed')) {
+            } else if (message.includes('Screenshot comparison failed') ||
```
sipayRT, line 215: we can merge these checks to your _isScreenshotComparisonError function

lib/adapters/test-result/playwright.ts
```diff
@@ -212,7 +212,8 @@ export class PlaywrightTestResultAdapter implements ReporterTestResult {
...
+                this._testResult.errors.every(error => this._isScreenshotComparisonError(error))) {
```
sipayRT: maybe you mean `some`, not `every`? we should mark error as ImageErro if at least one error will be with image diff

lib/adapters/test-result/playwright.ts
```diff
@@ -368,6 +369,23 @@ export class PlaywrightTestResultAdapter implements ReporterTestResult {
         return _.groupBy(imageAttachments, a => a.name.replace(ANY_IMAGE_ENDING_REGEXP, ''));
     }
 
+    private _isScreenshotComparisonError(error: PlaywrightTestResult['errors'][number]): boolean {
+        const message = stripAnsi(error.message || '');
+        const header = message.split('\n')[0];
```
sipayRT, line 374: splitting the message into the header and rest can lead to the search error. For example, in case with

```
await expect(page, 'Home page should match').toHaveScreenshot('custom.png');
```

header part will consist only text like "Error: Home page should match". In this case the test will be marked as none-image error

lib/adapters/test-result/playwright.ts
sipayRT: can you, pls, add one more unit with state name "some state"? pwt can replace a space with a dash here

